PR #2, opened 2022-11-02
feat: github repo setup

richardcase:
A few changes to setup the GitHub repo for future contribution. Includes:

- GHA workflow for labeling PRs based on size
- GHA workflow that requires a kind/* label on PRs
- GHA workflow for marking issues/prs as stale
- GHA workflow for running tests on PRs (i.e. build, test, verify)
- Dependabot configuration
- GitHub release notes configuration

Labels have also been created in the repo and these are used in the above.

Signed-off-by: Richard Case <richard.case@outlook.com>

PR #3, opened 2022-11-02
Standardise API imports

alexander-demicev:
Let's standardise API imports from the beginning so we don't miss them later.


richardcase:
Nice 👍 The linter should be happy after #2

richardcase (dismissed):

richardcase (approved):

PR #18, opened 2022-11-09
Adding first structs for the API

belgaied2:
This PR adds the first version of the structs `RKE2Config` for the provider's API.

Fixes #

This PR should be used as a support for  discussion and global requirements review. Let's discuss changes in the comments.

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests


bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	ImageCredentialProviderConfigMap ConfigMapRef `json:"imageCredentialProviderConfigMap,omitempty"`
+
+	// TODO: Remove ContainerRuntimeEndpoint since this feature will probably not be offered by CAPI Bootstrap provider?
+
+	// ContainerRuntimeEndpoint Disable embedded containerd and use alternative CRI implementation
+	// +optional
+	ContainerRuntimeEndpoint string `json:"containerRuntimeEndpoint,omitempty"`
+
+	// Snapshotter Override default containerd snapshotter (default: "overlayfs")
```
alexander-demicev, line 153: whare are the other options besides `overlayfs `? if they are limited, we can make an enum

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	// NodeExternalIp string `json:"nodeExternalIp,omitempty"`
+
+	// ResolvConf Kubelet resolv.conf file
+	// +optional
+	ResolvConf string `json:"resolvConf,omitempty"`
+
+	// ProtectKernelDefaults Kernel tuning behavior. If set, error if kernel tunables are different than kubelet defaults.
+	// +optional
+	ProtectKernelDefaults string `json:"protectKernelDefaults,omitempty"`
```
alexander-demicev, line 173: is this a bool flag?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+
+	// SystemDefaultRegistry Private registry to be used for all system images
+	// +optional
+	SystemDefaultRegistry string `json:"systemDefaultRegistry,omitempty"`
+
+	// Selinux Enable SELinux in containerd
+	// +optional
+	Selinux string `json:"selinux,omitempty"`
```
alexander-demicev, line 181:
```suggestion
	SELinux string `json:"selinux,omitempty"`
```

alexander-demicev: is this also a bool flag?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+
+	// LbServerPort Local port for supervisor client load-balancer. If the supervisor and apiserver are not colocated an additional port 1 less than this port will also be used for the apiserver client load-balancer. (default: 6444)
+	// +optional
+	LbServerPort string `json:"lbServerPort,omitempty"`
```
alexander-demicev, line 185:
```suggestion
	LBServerPort string `json:"lbServerPort,omitempty"`
```

alexander-demicev: what about making this field an int?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+
+	// Cni CNI Plugins to deploy, one of none, calico, canal, cilium; optionally with multus as the first value to enable the multus meta-plugin (default: canal)
+	// +optional
+	Cni Cni `json:"cni,omitempty"`
```
alexander-demicev, line 189:
```suggestion
	CNI CNI `json:"cni,omitempty"`
```

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+
+	// Selinux Enable SELinux in containerd
+	// +optional
+	Selinux string `json:"selinux,omitempty"`
+
+	// LbServerPort Local port for supervisor client load-balancer. If the supervisor and apiserver are not colocated an additional port 1 less than this port will also be used for the apiserver client load-balancer. (default: 6444)
+	// +optional
+	LbServerPort string `json:"lbServerPort,omitempty"`
+
+	// Cni CNI Plugins to deploy, one of none, calico, canal, cilium; optionally with multus as the first value to enable the multus meta-plugin (default: canal)
```
alexander-demicev, line 187: we can make this field an enum

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	Cni Cni `json:"cni,omitempty"`
+
+	// PauseImage Override image to use for pause
+	// +optional
+	PauseImage string `json:"pauseImage,omitempty"`
+
+	// RuntimeImage Override image to use for runtime binaries (containerd, kubectl, crictl, etc)
```
alexander-demicev, line 195: can this be an enum also?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	ControlPlaneResourceLimits string `json:"controlPlaneResourceLimits,omitempty"`
+
+	// Etcd defines optional custom configuration of ETCD
+	//+optional
+	Etcd ComponentConfig `json:"etcd,omitempty"`
+
+	// KubeApiServer defines optional custom configuration of the Kube API Server
+	//+optional
+	KubeApiServer ComponentConfig `json:"kubeApiServer,omitempty"`
```
alexander-demicev, line 234:
```suggestion
	KubeAPIServer ComponentConfig `json:"kubeApiServer,omitempty"`
```

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	//+optional
+	CloudControllerManager ComponentConfig `json:"cloudControllerManager,omitempty"`
+
+	// Kubelet defines optional custom configuration of the Kubelet
+	//+optional
+	Kubelet ComponentConfig `json:"kubelet,omitempty"`
+}
+
+type RKE2AgentConfig struct {
+	// DataDir Folder to hold state
```
alexander-demicev, line 258: I see a lot of duplications from the server config, what if we make a structure with a common field and have it nested in the server and client config?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	KubeProxy ComponentConfig `json:"kubeProxy,omitempty"`
+}
+
+// RKE2ConfigStatus defines the observed state of RKE2Config
+type RKE2ConfigStatus struct {
+	// Ready indicates the BootstrapData field is ready to be consumed
+	Ready bool `json:"ready,omitempty"`
+
+	BootstrapData []byte `json:"bootstrapData,omitempty"`
```
alexander-demicev, line 337: can we provider any description? also, why is this stored as a byte array?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+// +kubebuilder:validation:enum=rke2-coredns;rke2-ingress-nginx;rke2-metrics-server
+type DisabledItem string
+
+// CisProfile defines the CIS Benchmark profile to be activated in RKE2
+// +kubebuilder:validation:enum=cis-1.23
+type CisProfile string
+
+// Cni defines the Cni options for deploying RKE2
+// +kubebuilder:validation:enum=none;calico;canal;cilium
+type Cni string
```
alexander-demicev, line 399:
```suggestion
type CNI string
```

alexander-demicev:
```golang
const (
    Cillium CNI = "cillum"
)
```
we can create some constants to make it easier to manipulate the API programmatically.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	Permissions string `json:"permissions,omitempty"`
+
+	// Encoding specifies the encoding of the file contents.
+	// +optional
+	Encoding Encoding `json:"encoding,omitempty"`
+
+	// Content is the actual content of the file.
+	// +optional
+	Content string `json:"content,omitempty"`
```
alexander-demicev, line 433: is this field needed? what values is it supposed to store?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	ContentFrom *FileSource `json:"contentFrom,omitempty"`
+}
+
+// FileSource is a union of all possible external source types for file data.
+// Only one field may be populated in any given instance. Developers adding new
+// sources of data for target systems should add them here.
+type FileSource struct {
+	// Secret represents a secret that should populate this file.
+	Secret SecretFileSource `json:"secret"`
```
alexander-demicev, line 445: we should also consider adding a configmap as a source

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+}
+
+// Adapts a Secret into a FileSource.
+//
+// The contents of the target Secret's Data field will be presented
+// as files using the keys in the Data field as the file names.
+type SecretFileSource struct {
+	// Name of the secret in the RKE2BootstrapConfig's namespace to use.
+	Name string `json:"name"`
```
alexander-demicev, line 454: We can use localobjectrefence structure here, https://pkg.go.dev/k8s.io/api/core/v1#LocalObjectReference

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+
+	// Key is the key in the secret's data map for this value.
+	Key string `json:"key"`
+}
+
+type EtcdBackupConfig struct {
+	// EtcdDisableSnapshots Disable automatic etcd snapshots
+	// +optional
+	EtcdDisableSnapshots bool `json:"etcdDisableSnapshots,omitempty"`
```
alexander-demicev, line 463:
```suggestion
	DisableSnapshots bool `json:"disableSnapshots,omitempty"`
```

alexander-demicev: can we remove `Etcd` prefix from field name since it's already grouped under etcd structure.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	Mirrors map[string]Mirror `yaml:"mirrors" json:"mirrors,omitempty"`
+
+	// Configs are configs for each registry.
+	// The key is the FDQN or IP of the registry.
+	Configs map[string]RegistryConfig `yaml:"configs" json:"configs,omitempty"`
+}
+
+// Mirror contains the config related to the registry mirror
+type Mirror struct {
+
```
alexander-demicev, line 499:
```suggestion

```

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	TLS TLSConfig `yaml:"tls" json:"tls,omitempty"`
+}
+
+// AuthConfig contains the config related to authentication to a specific registry
+// This is not used in the CRDs but is used for marshalling the registries.yaml file
+type AuthConfig struct {
+	// Username is the username to login the registry.
+	Username string `yaml:"username" json:"username,omitempty"`
+	// Password is the password to login the registry.
+	Password string `yaml:"password" json:"password,omitempty"`
```
alexander-demicev, line 530: Let's use an objectrefence structure https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference for referencing a secret with username and password, and document that secret _must_ contain these keys.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+
+	// IdentityToken is used to authenticate the user and get
```
alexander-demicev, line 532: probably the same for the token

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+// - Used to generate a registries.yaml file for the RKE2 command line
+// In the CRD, CASecret attribute is a reference to a Secret in the Kube API, while in the registries.yaml CASecret is marshalled into ca_file which a file path on the system
+// This means the Registry object will need to be copied and processed before marshalling
+
+// TLSConfig contains the CA/Cert/Key used for a registry
+type TLSConfig struct {
+
+	// CASecret is a reference to a secret containing the Certificate Authority used to verify the registry's certificate
+	//+optional
+	CASecret SecretRef `json:"caSecret,omitempty"`
```
alexander-demicev, line 548: can CASecret, CertSecret, KeySecret be stored in one secret?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +required
+	Endpoint string `json:"endpoint"`
+
+	// EndpointCA references the Secret that contains a custom CA that should be trusted to connect to S3 endpoint
+	// +optional
+	EndpointCA SecretRef `json:"endpointCA,omitempty"`
+
+	// SkipSslVerify Disables S3 SSL certificate validation
+	// +optional
+	SkipSslVerify string `json:"skipSslVerify,omitempty"`
```
alexander-demicev, line 575: is this supposed to be a bool flag?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	//+optional
+	ExtraMounts map[string]string `json:"extraMounts,omitempty"`
+
+	//OverrideImage is a string that references a container image to override the default one for the Kubernetes Component
+	//+optional
+	OverrideImage string `json:"overrideImage,omitempty"`
+}
+
+// SecretRef references a secret using a namespace and a name
+type SecretRef struct {
```
alexander-demicev, line 614: this structure can be replaced with objectrefence structure https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference

alexander-demicev (commented):
it's great to see progress on the API, thanks a lot!

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	Files []File `json:"files,omitempty"`
 
-	// Foo is an example field of Rke2Config. Edit rke2config_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// PreRKE2Commands specifies extra commands to run before rke2 setup runs
+	// +optional
+	PreRKE2Commands []string `json:"preRKE2Commands,omitempty"`
+
+	// DeployRKE2Commands enables to override commands that are executed to setup rke2.
+	// It may be leveraged to deploy
```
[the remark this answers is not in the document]
belgaied2: Basically, it overrides the RKE2 install command, which supports a couple of options, but I just figured out that all the options are covered by the `RKE2ServerConfig `struct, which means, we don't need to expose the `DeployRKE2Commands `attribute to the user.

I will remove this attribute.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	//+optional
+	CertSecret SecretRef `json:"certSecret,omitempty"`
+
+	// KeySecret is a reference to a secret containing the Key for the certificate used to authenticate to the registry
+	//+optional
+	KeySecret SecretRef `json:"keySecret,omitempty"`
+
+	//InsecureSkipVerify may be set to true to skip verifying the registry's certificate
+	//+optional
+	InsecureSkipVerify bool `yaml:"insecure_skip_verify" json:"insecureSkipVerify,omitempty"`
```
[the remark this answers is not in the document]
belgaied2: I am ok with that, but I am wondering why.. is `bool` not recommended ?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: As of RKE2's documentation, you are right, this should be a `bool`. However, does the same idea as the above comment for `InsecureSkipVerify` also apply here?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	TLSSan []string `json:"tlsSan,omitempty"`
+
+	// DataDir is the Folder to hold RKE2's state (default: "/var/lib/rancher/rke2")
+	//+optional
+	DataDir string `json:"data-dir,omitempty"`
+
+	// ServiceNodePortRange is the port range to reserve for services with NodePort visibility (default: "30000-32767")
+	//+optional
+	ServiceNodePortRange string `json:"service-node-port-range,omitempty"`
```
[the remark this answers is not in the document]
belgaied2: Well, in that case, it would need to be 2 ints, one for the `Range Max` and one for the `Range Min`. But, the idea is not bad, it will need however some research on what are the possible values.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	ClusterDNS string `json:"clusterDNS,omitempty"`
+
+	// ClusterDomain Cluster Domain (default: "cluster.local")
+	// +optional
+	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// WriteKubeconfig gives the path to which kubeconfig file for admin client will be written
+	// +optional
+	WriteKubeconfig string `json:"writeKubeconfig,omitempty"`
```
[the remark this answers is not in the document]
belgaied2: Not really, it was just left there, but if we decide the user should not access the machine directly and should not want to change this field, we can remove it.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	//NodeIp string `json:"nodeIp,omitempty"`
+
+	// NodeExternalIp IPv4/IPv6 external IP addresses to advertise for node
+	// +optional
+	// NodeExternalIp string `json:"nodeExternalIp,omitempty"`
+
+	// ResolvConf Kubelet resolv.conf file
+	// +optional
+	ResolvConf string `json:"resolvConf,omitempty"`
```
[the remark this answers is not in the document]
belgaied2: This was supposed to be a path in RKE2 docs, and a file content in CABPR.
I suggest we change it into an ObjectReference to a ConfigMap.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: The list is available [here](https://github.com/containerd/containerd/blob/f033f6ff85ce397202ee69de22b4e3a4bf4093f5/docs/snapshotters/README.md#snapshotters), but the more I think about this, the more I am convinced we should not expose this option to users. 

I think it is way too advanced, and might cause more headaches than anything else. 

What do you think?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Good catch! Yes, it is supposed to be `bool`! Do you suggest an approach such as `ProtectKernelDefaultsPolicy` as an enum ?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Good catch again!

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Yes, I agree, but what about "nested anonymous struct" where struct fields are simply promoted to the upper struct, like the following:

```go
type RKE2ServerConfig struct {
        // Same attributes as Agent
        RKE2AgentConfig

        // BindAddress describes the rke2 bind address (default: 0.0.0.0).
	//+optional
	BindAddress string `json:"bindAddress,omitempty"`

	// AdvertiseAddress IP address that apiserver uses to advertise to members of the cluster (default: node-external-ip/node-ip).
	//+optional
	AdvertiseAddress string `json:"advertiseAddress,omitempty"`

	// TLSSan Add additional hostname or IP as a Subject Alternative Name in the TLS cert.
	//+optional
	TLSSan []string `json:"tlsSan,omitempty"`
(...)
```
We could also just suppose that for the specific case of CAPI, the user must define the same common settings for Agent and Server, meaning, we completely remove the Agent settings from the Server.

What do you think?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Yes, sure, however, the kubebuilder validation line should stay the same, right?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: It looks like this is inherited from K3s provider, I thought it did make sense but this code search suggests that this thing is unnecessary. Do you agree ?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: This is also inherited from the K3s Provider, it looks like the dev wanted to give the user the possibility to either reference content for a file or put in this field the `string` content inline.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: We could, but do we need the secret to be `local` ? We might just use a `v1.ObjectReference` instead.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Makes sense, I will take care of this!

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: The `AuthConfig` struct is not used for the API, and in the `RegistryConfig.AuthSecret` attribute, I do use a `v1.ObjectReference`, so I think we can resolve this item ?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: This could be, however, Kubernetes supports `TLS`-type secrets, and I thought I would suggest such a format for the user. Aggregating all certificates in one secret would need the secret to be a standard `Opaque `secret, but it is a trade-off

bootstrap/api/v1alpha1/rke2config_types.go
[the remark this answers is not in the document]
alexander-demicev: Oh, it's a range, I didn't notice. Let's keep the string for simplicity.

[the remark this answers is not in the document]
belgaied2: OK, I will resolve this conversation then!

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: yeah, we can remove it and add it back if someone asks for support.

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: yes, thanks!

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: Yeah, it looks unnecessary

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: yes, we are only adding constants to help people configure the provider from the code.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// DeployRKE2Commands enables to override commands that are executed to setup rke2.
+	// It may be leveraged to deploy
+	// +optional
+	DeployRKE2Commands []string `json:"deployRKE2Commands,omitempty"`
+
+	// PostRKE2Commands specifies extra commands to run after rke2 setup runs
+	// +optional
+	PostRKE2Commands []string `json:"postRKE2Commands,omitempty"`
+
+	// AgentConfig specifies configuration for the agent nodes
```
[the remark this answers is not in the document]
belgaied2: This should be solved by now.

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: +1, thanks a lot!

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// ImageCredentialProviderBinDir The path to the directory where credential provider plugin binaries are located (default: "/var/lib/rancher/credentialprovider/bin")
+	// +optional
+	//ImageCredentialProviderBinDir string `json:"imageCredentialProviderBinDir,omitempty"`
+
+	// ImageCredentialProviderConfigMap is a reference to the ConfigMap that contains credential provider plugin config
+	// The configMap should contain a YAML file content + a Path to the Binaries for Credential Provider.
+	// +optional
+	ImageCredentialProviderConfigMap ConfigMapRef `json:"imageCredentialProviderConfigMap,omitempty"`
+
+	// TODO: Remove ContainerRuntimeEndpoint since this feature will probably not be offered by CAPI Bootstrap provider?
```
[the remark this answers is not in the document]
belgaied2: This is Kubelet feature since Kubernetes v1.20, official documentation [here](https://kubernetes.io/docs/tasks/kubelet-credential-provider/kubelet-credential-provider/).

bootstrap/api/v1alpha1/rke2config_types.go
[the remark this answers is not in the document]
belgaied2: Done, but not yet in the PR. Does Kubebuilder offer an easy way to validate that the referenced object should be a ConfigMap ?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,593 @@ limitations under the License.
...
+	//+optional
+	ServerConfig RKE2ServerConfig `json:"serverConfig,omitempty"`
+
+	// PrivateRegistriesConfig defines the containerd configuration for private registries and local registry mirrors.
+	//+optional
+	PrivateRegistriesConfig Registry `json:"privateRegistriesConfig,omitempty"`
+
+	// Version specifies the rke2 version.
+	//+optional
+	Version string `json:"version,omitempty"`
```
richardcase, line 56: What would we use this version for? Would it be used to download rke2 if its not on the base image?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Done, using `SSLVerifyPolicy` as a reusable enum type for both S3 backups and Private Registry configrations.

bootstrap/api/v1alpha1/rke2config_types.go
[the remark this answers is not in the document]
belgaied2: done

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Done, also for replacing `bool` with enum

bootstrap/api/v1alpha1/rke2config_types.go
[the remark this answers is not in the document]
richardcase: Yep we should remove anything related to kubeconfig

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,593 @@ limitations under the License.
...
+	//+optional
+	EtcdBackupConfig EtcdBackupConfig `json:"etcdBackupConfig,omitempty"`
+
+	// Disable Do not deploy packaged components and delete any deployed components (valid items: rke2-coredns, rke2-ingress-nginx, rke2-metrics-server).
+	//+optional
+	Disable []DisabledItem `json:"disable,omitempty"`
+
+	// DisableScheduler Disable Kubernetes default scheduler.
+	//+optional
+	DisableScheduler string `json:"disable-scheduler,omitempty"`
```
richardcase, line 113: With this, `DisableCloudController`, `DisableKubeProxy` we can add them to their respective structs.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	WriteKubeconfig string `json:"writeKubeconfig,omitempty"`
+
+	// WriteKubeconfigMode Writes kubeconfig with this mode
+	// +optional
+	WriteKubeconfigMode string `json:"writeKubeconfigMode,omitempty"`
+
+	// EtcdExposeMetrics Expose etcd metrics to client interface. (Default false)
+	// +optional
+	EtcdExposeMetrics string `json:"etcdExposeMetrics,omitempty"`
```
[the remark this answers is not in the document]
belgaied2: Done

bootstrap/api/v1alpha1/rke2config_types.go
richardcase: Or just `LoadBalancerPort`?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,593 @@ limitations under the License.
...
+	//+optional
+	CisProfile CisProfile `json:"cisProfile,omitempty"`
+
+	// AuditPolicySecret Path to the file that defines the audit policy configuration.
+	//+optional
+	AuditPolicySecret corev1.ObjectReference `json:"auditPolicySecret,omitempty"`
+
+	// ControlPlaneResourceRequests Control Plane resource requests.
+	//+optional
+	ControlPlaneResourceRequests string `json:"controlPlaneResourceRequests,omitempty"`
```
richardcase, line 220: Move to a control plane struct?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Done!

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Done!

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Done!

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: If you are mentioning the Cni attribute, it is an enum with Kubebuilder validation on type (not on field). Is there a reason I should do it differently ?

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,593 @@ limitations under the License.
...
+}
+
+type RKE2AgentConfig struct {
+	// DataDir Folder to hold state.
+	//+optional
+	DataDir string `json:"dataDir,omitempty"`
+
+	// NodeLabels  Registering and starting kubelet with set of labels.
+	//+optional
+	NodeLabels []string `json:"nodeLabels,omitempty"`
```
richardcase, line 262: There seems to be some fields in the RKE2AgentConfig struct that are the same from the RKE2ServerConfig struct. We could put these into a common struct and inline it in both RKE2AgentConfig and RKE2ServerConfig

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: If you are mentioning the RuntimeImage field, it cannot be an `enum` because it is a custom image name, and the image will contain components such as containerd, kubectl, etc.

bootstrap/api/v1alpha1/rke2config_types.go
richardcase: This is the same in the kubeadm bootstrap provider

bootstrap/api/v1alpha1/rke2config_types.go
richardcase: This would make it different from CAPBPK

richardcase (commented):
Only a very initial review. I will have a deeper look later.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Acted to use "nested anonymous struct"

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// +optional
+	EtcdExposeMetrics string `json:"etcdExposeMetrics,omitempty"`
+
+	// EtcdBackupConfig defines how RKE2 will snapshot ETCD: target storage, schedule, etc.
+	//+optional
+	EtcdBackupConfig EtcdBackupConfig `json:"etcdBackupConfig,omitempty"`
+
+	// Disable Do not deploy packaged components and delete any deployed components (valid items: rke2-coredns, rke2-ingress-nginx, rke2-metrics-server)
+	// +optional
+	Disable []DisabledItem `json:"disable,omitempty"`
```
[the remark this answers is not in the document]
belgaied2: Acted on creating a struct called `Disable` with a sub-struct `KubernetesComponents` and another `Apps` ...

bootstrap/api/v1alpha1/rke2config_types.go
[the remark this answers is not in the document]
belgaied2: Acted on removing kubeconfig options.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Yes, the idea was to do that. RKE2 accepts an environment variable when running its installation script that specifies the version.

It might be possible to get the version from the the `Cluster` object for example, but the exact RKE2 version is slightly different than Kubernetes. If Kubernetes has for example a version  `v1.25.5`, RKE2 might have a version name such as `v1.25.5+rke2r1` or `v1.25.5+rke2r2`.

alexander-demicev: Reading it from `Cluster` should be fine, even if RKE2 versions are different.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,525 @@ limitations under the License.
...
+
+	// PauseImage Override image to use for pause.
+	//+optional
+	PauseImage string `json:"pauseImage,omitempty"`
+
+	// RuntimeImage Override image to use for runtime binaries (containerd, kubectl, crictl, etc).
+	//+optional
+	RuntimeImage string `json:"runtimeImage,omitempty"`
+
+	//	CloudProviderName  Cloud provider name.
```
alexander-demicev, line 118:
```suggestion
	// CloudProviderName  Cloud provider name.
```

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,525 @@ limitations under the License.
...
+	//+optional
+	CloudProviderName string `json:"cloudProviderName,omitempty"`
+
+	//	CloudProviderConfigMap  is a reference to a ConfigMap containing Cloud provider configuration.
```
alexander-demicev, line 122:
```suggestion
	// CloudProviderConfigMap  is a reference to a ConfigMap containing Cloud provider configuration.
```

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,525 @@ limitations under the License.
...
+type ComponentConfig struct {
+	// ExtraEnv is a map of environment variables to pass on to a Kubernetes Component command.
+	//+optional
+	ExtraEnv map[string]string `json:"extraEnv,omitempty"`
+
+	// ExtraArgs is a map of command line arguments to pass to a Kubernetes Component command.
+	//+optional
+	ExtraArgs map[string]string `json:"extraArgs,omitempty"`
+
+	//ExtraMounts is a map of volume mounts to be added for the Kubernetes component StaticPod
```
alexander-demicev, line 530:
```suggestion
	// ExtraMounts is a map of volume mounts to be added for the Kubernetes component StaticPod
```

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,525 @@ limitations under the License.
...
+	//+optional
+	ExtraMounts map[string]string `json:"extraMounts,omitempty"`
+
+	//OverrideImage is a string that references a container image to override the default one for the Kubernetes Component
```
alexander-demicev, line 534:
```suggestion
	// OverrideImage is a string that references a container image to override the default one for the Kubernetes Component
```

alexander-demicev (commented):
left some more comments, awesome to see the API getting better. Thanks @belgaied2

controlplane/api/v1alpha1/rke2controlplane_types.go
```diff
@@ -17,48 +17,270 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"github.com/rancher-sandbox/cluster-api-provider-rke2/bootstrap/api/v1alpha1"
```
alexander-demicev, line 20:
```suggestion
	bootstrapv1 "github.com/rancher-sandbox/cluster-api-provider-rke2/bootstrap/api/v1alpha1"
```

controlplane/api/v1alpha1/rke2controlplane_types.go
```diff
@@ -17,48 +17,270 @@ limitations under the License.
...
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
alexander-demicev, line 25: we can remove this comment

controlplane/api/v1alpha1/rke2controlplane_types.go
```diff
@@ -17,48 +17,270 @@ limitations under the License.
...
-// Rke2ControlPlaneSpec defines the desired state of Rke2ControlPlane
-type Rke2ControlPlaneSpec struct {
+// RKE2ControlPlaneSpec defines the desired state of RKE2ControlPlane
+type RKE2ControlPlaneSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// Foo is an example field of Rke2ControlPlane. Edit rke2controlplane_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// ServerConfig specifies configuration for the agent nodes.
```
richardcase, line 33: I think we need to have a field of type `bootstrapv1.RKE2ConfigSpec` in this struct

richardcase: Or we inline the struct instead of using a struct field. Either way we should include it.

bootstrap/api/v1alpha1/rke2config_types.go
richardcase: If its for the Kubernetes version then its fine to keep it here, as long as we are importing this inlining/using this struct in the control plane api definition

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,314 @@ limitations under the License.
...
+	Version string `json:"version,omitempty"`
+}
+
+type RKE2AgentConfig struct {
+	// RKE2CommonNodeConfig is an inline struct that references common attribtes between agent and server nodes
+	RKE2CommonNodeConfig `json:",inline"`
+}
+
+// RKE2CommonNodeConfig describes some attributes that are common to agent and server nodes
+type RKE2CommonNodeConfig struct {
```
richardcase, line 61: Might be good to extract this struct definition to a `types.go` file in this package. So that we keep this file directly for rke2config types.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -17,48 +17,314 @@ limitations under the License.
...
+	// Path specifies the full path on disk where to store the file.
+	Path string `json:"path"`
+
+	// Owner specifies the ownership of the file, e.g. "root:root".
+	//+optional
+	Owner string `json:"owner,omitempty"`
+
+	// Permissions specifies the permissions to assign to the file, e.g. "0640".
+	//+optional
+	Permissions string `json:"permissions,omitempty"`
```
richardcase, line 237: What format of string is expected here? Could we add a validation regex?

bootstrap/api/v1alpha1/rke2configtemplate_webhook.go
```diff
@@ -36,39 +36,39 @@ func (r *Rke2ConfigTemplate) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
 // Default implements webhook.Defaulter so a webhook will be registered for the type
-func (r *Rke2ConfigTemplate) Default() {
-	rke2configtemplatelog.Info("default", "name", r.Name)
+func (r *RKE2ConfigTemplate) Default() {
+	RKE2configtemplatelog.Info("default", "name", r.Name)
 
 	// TODO(user): fill in your defaulting logic.
 }
 
 // TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
```
richardcase, line 48: nit, we can remove these todos that were left in as part of scaffolding. But this can be done later.

controlplane/api/v1alpha1/rke2controlplane_types.go
richardcase: Or we move fields like version, files, PreRKE2Commands to the common struct.

controlplane/api/v1alpha1/rke2controlplane_types.go
```diff
@@ -17,48 +17,270 @@ limitations under the License.
...
+	// Each data entry in the ConfigMap will be will be copied to a folder on the control plane nodes that RKE2 scans and uses to deploy manifests.
+	//+optional
+	ManifestsConfigMapReference corev1.ObjectReference `json:"manifestsConfigMapReference,omitempty"`
 }
 
-// Rke2ControlPlaneStatus defines the observed state of Rke2ControlPlane
-type Rke2ControlPlaneStatus struct {
+type RKE2ServerConfig struct {
+	// RKE2CommonNodeConfig is an inline struct that references common attribtes between agent and server nodes
+	v1alpha1.RKE2CommonNodeConfig `json:",inline"`
```
richardcase, line 45: If we use `bootstrapv1.RKE2ConfigSpec`  directly in RKE2ControlPlaneSpec then this could be removed from here (see previous comment first)

controlplane/api/v1alpha1/rke2controlplane_types.go
```diff
@@ -17,48 +17,270 @@ limitations under the License.
...
+	//+optional
+	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// DisableComponents lists Kubernetes components and RKE2 plugin components that will be disabled.
+	//+optional
+	DisableComponents DisableComponents `json:"disableComponents,omitempty"`
+
+	// LoadBalancerPort Local port for supervisor client load-balancer. If the supervisor and apiserver are not colocated an additional port 1 less than this port will also be used for the apiserver client load-balancer (default: 6444).
+	//+optional
+	LoadBalancerPort int `json:"lbServerPort,omitempty"`
```
richardcase, line 77: json tag name is different, they should be the same except formatting

controlplane/api/v1alpha1/rke2controlplane_types.go
```diff
@@ -17,48 +17,270 @@ limitations under the License.
...
+
+	// NOTE: this was only profile, changed it to cisProfile.
+
+	// AuditPolicySecret Path to the file that defines the audit policy configuration.
+	//+optional
+	AuditPolicySecret corev1.ObjectReference `json:"auditPolicySecret,omitempty"`
+
+	// ControlPlaneResourceRequests Control Plane resource requests.
+	//+optional
+	ControlPlaneResourceRequests string `json:"controlPlaneResourceRequests,omitempty"`
```
richardcase, line 108: The requests and limits would be on the Deployment spec for the providers controller manager. We could probably remove these.

richardcase (commented):
Looking good.

I think we should think about merging this change in soon. We can then start the reconciliation implementation and make any API changes as we go.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: Yes, definitely.

This part comes BTW from the `Kthrees` provider, which creates a `write_files` primitive in Cloud-init. That is documented [here](https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=write_files#write-files).

Basically, it is an octal string for file permissions, either `'0644' or '0o644'` for a file that is read-write for the owner, read-only for all others, without execution rights. Validation would probably be as such:
- Length can be 4 or 5
- First character should be `0`
- If Length is 4, then each character could be any digit between `0` and `7`.
- If Length is 5, then second character should be 'o' and all other characters should be any digit between `0` and `7`.

So something like this:
`0(|o)[0-7]{3}`

controlplane/api/v1alpha1/rke2controlplane_types.go
belgaied2: Good catch! I will change that!

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: If we have 2 versions of RKE2 for a Kubernetes version, how do we know which exact version the user wanted?
We might for example query GitHub for the latest release of RKE2 for that particular Kubernetes version (let's say v1.25.5+rke2r2), but what if the user wants v1.25.5+rke2r1 because that's what's on the Rancher Support Matrix ?

So, I would be of the opinion that it is more a specific RKE2 version, not only a Kubernetes version. It might however need validation on the Controller level.

controlplane/api/v1alpha1/rke2controlplane_types.go
belgaied2: Done!

controlplane/api/v1alpha1/rke2controlplane_types.go
belgaied2: Done!

controlplane/api/v1alpha1/rke2controlplane_types.go
belgaied2: Done!

controlplane/api/v1alpha1/rke2controlplane_types.go
belgaied2: Done!

belgaied2 (commented):
Please, redo a short review.

bootstrap/api/v1alpha1/rke2configtemplate_webhook.go
belgaied2: Done!

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -18,47 +18,624 @@ package v1alpha1
...
+	// Region S3 region / bucket location (optional) (default: "us-east-1")
+	// +optional
+	Region string `json:"region,omitempty"`
+
+	// Folder S3 folder
+	// +optional
+	Folder string `json:"folder,omitempty"`
+}
+
+type ComponentConfig struct {
```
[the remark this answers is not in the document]
alexander-demicev: It's fine to leave it here

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: I would assume that the version field for our provider always means an RKE2 version.

alexander-demicev (dismissed):
great work @belgaied2! let's improve the API from here in follow-up PRs.

bootstrap/api/v1alpha1/rke2config_types.go
richardcase: Interesting, i would not make that assumption myself.

richardcase (approved):

PR #52, opened 2022-12-01
Add .golangci.yml config & golangci-lint.yml github action

### LINKED ISSUE #11: Configure golangci-lint
Currently we have no golangci-lint configuration. We should look to enable all linters and the explicitly exclude the ones we don't want.

aniruddha2000:
Signed-off-by: Aniruddha Basak <codewithaniruddha@gmail.com>

kind/feature

This PR contains the golangci-lint config & necessary GitHub actions so whenever a new PR open we have a clear picture of where the lint is falling and we also follow best practices while writing code for future readability of the project.

Fixes #11 

It's just an initial phase I added and I need some reviews from the community where I can improve the PR.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,27 @@
+name: golangci-lint
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+# Remove all permissions from GITHUB_TOKEN except metadata.
+permissions: {}
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        working-directory:
+          - ""
+    steps:
+      - uses: actions/checkout@v3.1.0
+      - uses: actions/setup-go@v3.3.1
+        with:
+          go-version: 1.19
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3.3.1
+        with:
+          version: v1.50.1
+          working-directory: ${{matrix.working-directory}}
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,13 @@
+# Options for analysis running.
+run:
+  go: "1.19"
+  skip-files:
+    - "zz_generated.*\\.go$"
+    - "vendored_openapi\\.go$"
+  allow-parallel-runners: true
+linters:
+  enable-all: true
+issues:
+  include:
+    - EXC0012
+    - EXC0014
--- a/Makefile
+++ b/Makefile
@@ -541,3 +541,8 @@ $(GH): # Download GitHub cli into the tools bin folder
 	hack/ensure-gh.sh \
 		-b $(TOOLS_BIN_DIR) \
 		$(GH_VERSION)
+
+$(GOLANGCI_LINT): .github/workflows/golangci-lint.yml # Download golangci-lint using hack script into tools folder.
+	hack/ensure-golangci-lint.sh \
+		-b $(TOOLS_BIN_DIR) \
+		$(shell cat .github/workflows/golangci-lint.yml | grep [[:space:]]version | sed 's/.*version: //')


.golangci.yml
```diff
@@ -0,0 +1,55 @@
+# Options for analysis running.
+run:
+  go: "1.19"
+  skip-files:
+    - "zz_generated.*\\.go$"
+    - "vendored_openapi\\.go$"
+  allow-parallel-runners: true
+linters:
+  disable-all: true
```
richardcase, line 9: I think we should flip it the other way. So `enable-all` and then explicitly disable linters we don't want. That way it forces us to consider new linters and whether we can them.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+      - uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: 1.19
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@0ad9a0988b3973e851ab0a07adf248ec2e100376 # tag=v3.3.1
+        with:
+          version: v1.50.1
+          working-directory: ${{matrix.working-directory}}
```
richardcase, line 29: Do we need a matrix here? Could we just run it over all the source?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        working-directory:
+          - ""
+          - test
+          - hack/tools
+    steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
```
richardcase, line 21: I'd favour using the tag instead of the commit hash. Same comment applies to the other actions

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: 1.19
+      - name: golangci-lint
```
richardcase, line 25: We have a `lint` target in ourt Makefile. It may be better to invoke that instead of calling golangci-lint directly in the action. wdyt?

richardcase (changes requested):
Thanks for this change @aniruddha2000.

.golangci.yml
alexander-demicev: on the other hand, we will be using the same linter rules as CAPI https://github.com/kubernetes-sigs/cluster-api/blob/main/.golangci.yml#L13 this way

aniruddha2000: Make sense

.github/workflows/golangci-lint.yml
aniruddha2000: Yes, then I should remove these lines with 
```yml
strategy:
      matrix:
        working-directory:
          - ""
          - test
          - hack/tools
```

as well right?

.github/workflows/golangci-lint.yml
aniruddha2000: Yes true

.github/workflows/golangci-lint.yml
aniruddha2000: Does that use binary?

In the [golang-ci](https://golangci-lint.run/usage/install/#github-actions) doc it's mentioned that it's recommended to use the GitHub action GitHub projects due to some smart caching it implements inside it that makes it fast.

.golangci.yml
aniruddha2000: Yes @alexander-demicev initially I took the idea from CAPI templates

richardcase: I don't think we need to follow CAPI here, as a provider, we can choose what linters to enable. It's far better to get the benefit of new linters and then have to make an explicit decision to disable one. If you use the disable all then you are not forced to evaluate new linters.

aniruddha2000: Resolving this as I have changed to enable all.

.github/workflows/golangci-lint.yml
richardcase: If we wanted to use the action then we could use the same approach as upstream CAPI to make sure the the make target (when run locally) uses the same version of golanci-lint: https://github.com/kubernetes-sigs/cluster-api/blob/main/Makefile#L1200:L1203

.github/workflows/golangci-lint.yml
richardcase: Lets leave it as is

aniruddha2000: Okay then resolving this

.github/workflows/golangci-lint.yml
aniruddha2000: Make sense

aniruddha2000:
@richardcase I think the lint has triggered.

Are we going to fix these lints in some future issues?

richardcase:
Wow, we have plenty of linter issues 🤣

aniruddha2000:
what should be the next step then 😹

richardcase:
I would say we merge with the lint failures and then follow up with changes and disabling a select few linters (along with the reasons for disabling.....perhaps in a ADR).

aniruddha2000:
Sounds Good.
So let's merge it. And if you are creating a new issue about disabling some linters and fixing the issues then I would like to work on that too. Also, I have noted down some linters that could be disabled.

aniruddha2000:
@richardcase should we merge it then?

richardcase (approved):
As discussed we are going to merge and fix post merge

PR #107, opened 2023-03-10
fixed linting issues (first pass)

belgaied2:
Linting introduced in PR #52 was showing a lot of errors. This PR aims at fixing that by:
- By fixing the code itself for many linters
- By changing the configuration of `.golangci.yaml` for issues deemed unimportant or non urgent. 

In this PR, there is a lot of changed files caused by fixing linting issues, mostly minor ones for lisibility of code such as: 
- wsl (empty lines)
- lll (long lines)
- gci (import organisation)
- gofumpt
- etc.

Some of the more impacting linters (funlen for instance or gosec etc.) were disabled because they would mean a larger effort, to define on a case-by-case basis.

fixed some regressions from refactoring. After this PR, functionality should be equal to v0.1.0-alpha.1.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


richardcase (approved):
Thats quite a long list of disabled linters but we can slowly re-enable over time so that task isn't too painful.

PR #110, opened 2023-03-14
Fixed #28 CIS Profile now works and is different for v1.25+ and v1.24-

### LINKED ISSUE #28: Handle CISProfile field correctly depending on RKE2's version
As of defined in the API, the `CISProfile` field can only take the value `cis-1.23` which RKE2's only CIS Profile possible in `v1.25`. In previous version of RKE2, the profile could be either `v1.5` or `v1.6` which relied on `PodSecurityPolicies`. PSPs are deprecated in Kubernetes since v1.21, and were dropped from the API in v1.25. 

The question here is:
- Should CAPBR support the older flags `v1.5` and `v1.6` because users might want to deploy an older version of RKE2 than `v1.25` or should it be forward looking by only supporting `v1.23`? In which case, the controller needs to make sure that the desired RKE2 version should be `>= v1.25` 
- If the decision is "yes, we support these older flags", the API will need to be updated.
- If the decision is "no, we will not support them", the API will not change but the controller needs to implement some kind of validation against RKE2's version.

Because there is a risk in the implementation to not account for compatibility with RKE2's version.

belgaied2:

This PR better implements the CIS Profile by differenciation RKE2 version that are <v1.25 which need a profile `cis-1.5` or `cis-1.6` and the versions >=v1.25, which need a profile `cis-1.23` replacing Pod Security Policies with Pod Security Admission.

It also fixes an issue where the node was not correctly prepared before RKE2 service starts, causing a failure during RKE2's installation.

Fixes #28

**Checklist**:

- [ ] includes documentation
- [X] adds unit tests

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -24,12 +24,19 @@ import (
 	"regexp"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/version"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 
+	bootstrapv1 "github.com/rancher-sandbox/cluster-api-provider-rke2/bootstrap/api/v1alpha1"
 	controlplanev1 "github.com/rancher-sandbox/cluster-api-provider-rke2/controlplane/api/v1alpha1"
+)
+
+const (
+	// RKE2_CIS_VERSION_CHANGE is the version where the CIS benchmark changed in RKE2 (because of PSPs).
+	RKE2_CIS_VERSION_CHANGE = "v1.25.0"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/util/util.go
```diff
@@ -161,3 +168,37 @@ func GetMapKeysAsString(m map[string][]byte) (keys string) {
 
 	return
 }
+
+// AtLeast_v1_25 returns true if the RKE2 version is at least v1.25.0.
+func AtLeast_v1_25(rke2version string) (bool, error) {
```
alexander-demicev, line 173:
```suggestion
func AtLeastv125(rke2version string) (bool, error) {
```
or 
```suggestion
func greaterThanV125(rke2version string) (bool, error) {
```

Snake case is not idiomatic in golang

alexander-demicev (changes requested):

pkg/util/util.go
belgaied2: First suggestion implemented

alexander-demicev (approved):

PR #114, opened 2023-03-21
implemented additionCloudInit configuration in the API

### LINKED ISSUE #78: Provide an API option to pass any cloudinit parameter.

cloud-init has a many features, and it would be painfull (and wouldn't add much value) to map each of these feature options in bootstrap provider API.

For example, RKE2 boostrap provider already supports File injection, and NTP support is planned. Kubeadm also supports disk-setup (which [would be also be desirable for this provider](https://github.com/rancher-sandbox/cluster-api-provider-rke2/issues/77)) and user-definitions. In bootstrap providers, these values are often passed as is to cloud init file, but they require a fair amount of boilerplate code.

The provider could instead propose an untyped option where the user would provide a map of cloud-init instructions. These instruction could be merged with ones that are generated by bootstrap provider ( https://github.com/rancher-sandbox/cluster-api-provider-rke2/issues/68 could help a lot on that front)

For some cloud-init instructions that are used by boostrap provider (like runCmd), there won't be any obvious way to merge them with user-provided ones. In that case, an admission webhook could be used to evict such user values and instruct him to use Pre/PostRKE2Commands instead.

belgaied2:

Fixes #78

**Checklist**:

- [ ] includes documentation
- [X] adds unit tests

### DIFF
--- a/bootstrap/api/v1alpha1/rke2config_types.go
+++ b/bootstrap/api/v1alpha1/rke2config_types.go
@@ -134,6 +134,11 @@ type RKE2AgentConfig struct {
 	// AirGapped is a boolean value to define if the bootstrapping should be air-gapped,
 	// basically supposing that online container registries and RKE2 install scripts are not reachable.
 	AirGapped bool `json:"airGapped,omitempty"`
+
+	// AdditionalCloudInit is a field that allows users to specify additional cloud-init configuration to be included in the
+	// generated cloud-init script.
+	//+optional
+	AdditionalCloudInit string `json:"additionalCloudInit,omitempty"`
 }
 
 // NTP defines input for generated ntp in cloud-init.
--- a/bootstrap/api/v1alpha1/rke2config_types.go
+++ b/bootstrap/api/v1alpha1/rke2config_types.go
@@ -135,10 +135,11 @@ type RKE2AgentConfig struct {
 	// basically supposing that online container registries and RKE2 install scripts are not reachable.
 	AirGapped bool `json:"airGapped,omitempty"`
 
-	// AdditionalCloudInit is a field that allows users to specify additional cloud-init configuration to be included in the
+	// AdditionalUserData is a field that allows users to specify additional cloud-init configuration to be included in the
 	// generated cloud-init script.
+	// NOTE: All fields of the UserData that are managed by the RKE2Config controller will be ignored, this include "write_files", "runcmd", "ntp".
 	//+optional
-	AdditionalCloudInit string `json:"additionalCloudInit,omitempty"`
+	AdditionalUserData string `json:"additionalUserData,omitempty"`
 }
 
 // NTP defines input for generated ntp in cloud-init.
--- a/bootstrap/internal/cloudinit/cloudinit.go
+++ b/bootstrap/internal/cloudinit/cloudinit.go
@@ -19,6 +19,7 @@ package cloudinit
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 	"text/template"
 
@@ -28,9 +29,15 @@ import (
 	bootstrapv1 "github.com/rancher-sandbox/cluster-api-provider-rke2/bootstrap/api/v1alpha1"
 )
 
-var defaultTemplateFuncMap = template.FuncMap{
-	"Indent": templateYAMLIndent,
-}
+var (
+	// defaultTemplateFuncMap is the default set of functions for the template.
+	defaultTemplateFuncMap = template.FuncMap{
+		"Indent": templateYAMLIndent,
+	}
+
+	// ignoredCloudInitFields is a list of fields that are ignored from additionalCloudInit when generating final configuration.
+	ignoredCloudInitFields = []string{"runcmd", "write_files", "ntp"}
+)
 
 func templateYAMLIndent(i int, input string) string {
 	split := strings.Split(input, "\n")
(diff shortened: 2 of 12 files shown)


bootstrap/internal/cloudinit/cloudinit.go
```diff
@@ -121,3 +124,37 @@ func generate(kind string, tpl string, data interface{}) ([]byte, error) {
 
 	return out.Bytes(), nil
 }
+
+func cleanupAdditionalCloudInit(cloudInitData string) (string, error) {
+	m := make(map[string]interface{})
+
+	err := yaml.Unmarshal([]byte(cloudInitData), m)
+	if err != nil {
+		return "", err
```
alexander-demicev, line 133: is it possible to add some context to errors here with `fmt.Errorf`, I expect at some point someone will pass an invalid yaml, and it's better to let users know more about what happened.

bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -134,6 +134,11 @@ type RKE2AgentConfig struct {
 	// AirGapped is a boolean value to define if the bootstrapping should be air-gapped,
 	// basically supposing that online container registries and RKE2 install scripts are not reachable.
 	AirGapped bool `json:"airGapped,omitempty"`
+
+	// AdditionalCloudInit is a field that allows users to specify additional cloud-init configuration to be included in the
+	// generated cloud-init script.
+	//+optional
+	AdditionalCloudInit string `json:"additionalCloudInit,omitempty"`
```
richardcase, line 141: I think we should probably choose a different name for this to make it a bit more generic. The reason being is that when we add support for Ignition we may want to allow a similar functionality to provide additional configuration. Perhaps something like **AdditionalBootstrapData** or something like that.

bootstrap/internal/cloudinit/cloudinit.go
```diff
@@ -121,3 +124,37 @@ func generate(kind string, tpl string, data interface{}) ([]byte, error) {
...
+	}
+
+	// Remove the runcmd section
+	delete(m, "runcmd")
```
richardcase, line 137: Instead of doing delete like this could we store the fields we want to remove in a slide and then iterate over it?

bootstrap/internal/cloudinit/cloudinit.go
```diff
@@ -121,3 +124,37 @@ func generate(kind string, tpl string, data interface{}) ([]byte, error) {
...
+
+	// Remove the write_files section
+	delete(m, "write_files")
```
richardcase, line 140: Also, do we want to create a warning that the extra bootstrap data contained a field that is going to be ignored?

richardcase: Or check in the validating webhook and reject the change?

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: @alexander-demicev the thing is : that `AdditionalConfig` field in Kubeadm is already inside a `IgnitionSpec` struct, so it is not generic, it is ignition specific. I thought I would be more specific to avoid misunderstandings.

bootstrap/internal/cloudinit/cloudinit.go
belgaied2: a warning might be useful, however, I don't think we should block the incoming bootstrap object just because there are field that will be ignored, it would be a hassle for the users IMHO.

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: @richardcase usually, this would be referred to as `userData` maybe I should put `AdditionalUserData`

bootstrap/internal/cloudinit/cloudinit.go
belgaied2: done!

bootstrap/internal/cloudinit/cloudinit.go
belgaied2: Done!

bootstrap/internal/cloudinit/cloudinit.go
```diff
@@ -121,3 +124,37 @@ func generate(kind string, tpl string, data interface{}) ([]byte, error) {
 
 	return out.Bytes(), nil
 }
+
+func cleanupAdditionalCloudInit(cloudInitData string) (string, error) {
+	m := make(map[string]interface{})
+
+	err := yaml.Unmarshal([]byte(cloudInitData), m)
```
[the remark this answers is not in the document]
belgaied2: Done!

bootstrap/api/v1alpha1/rke2config_types.go
alexander-demicev: in that case, it makes sense to name it `AdditionalCloudInit` so people don't confuse it with IgnitionSpec. Alternatively, we can add a new structure  `AdditionalUserData` that contains 2 fields inside: `IgnitionConfig`, `CloudInitConfig`. Why is API design so hard? :) I would prefer to avoid having to specify additional cloud-init config in one place and ignition in another nested object.

belgaied2: I suggest I change it to `AdditionalUserData` for now, and we can think about the details of ignition later. What do you think ?

bootstrap/internal/cloudinit/cloudinit.go
belgaied2: I documented in the API description that some fields would be ignored. Adding a warning involves some changes to the code that do not really make sense because the logger is not available in this nested func. What do you think ?

alexander-demicev (approved):

bootstrap/api/v1alpha1/rke2config_types.go
belgaied2: If I understand it right, we might ignore the way Kubeadm does it, and do it this way:
- Do not use a nested config for Ignition
- Use `AdditionalUserData` as a free text field which might contain Ignition or Cloud-Init configuration
- Figure out a way to make the user understand what he is doing (an enum field for choosing between cloud-init or ignition?)
- We might want to validate the `AdditionalUserData` field by parsing it for signs of ignition or cloud-init.

richardcase: I like this idea, that way we don't need 2 different fieldsfor the same thing

PR #115, opened 2023-03-21
moved useful constants to own package

### LINKED ISSUE #81: Move shared constants like `"root:root"` to a separate package
For example, in [this](https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/a8c0d89d98eb151488f19e36aa2cce346f654e49/pkg/rke2/config.go#L341) part of the code, we are using constants that we defined somewhere else. These constants can be moved to a shared package, for instance called `consts`, which has all shared constants.

It centralises constant values, decreasing maintainability effort for these constants if they ever need to change. It is also less error prone.

belgaied2:

This PR moves some constants and hard-coded values into public constants, in order to have a single reference for their values.

Fixes #81

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


alexander-demicev (approved):
That makes sense, thanks a lot for cleaning it up!

PR #116, opened 2023-03-21
Fixed a mistake in the description of ExtraArgs field

### LINKED ISSUE #112: Description of field ExtraArgs in Kubernetes components config says it should be a map whereas it should be a list
**What happened:**
ExtraArgs field should be a list not a map.

**How to reproduce it:**

**Environment:**

- rke provider version:

belgaied2:

The comment on the ExtraArgs field in ComponentConfig struct in the Bootstrap API suggests that the field should be a map whereas the field is actually a slice of strings with the format `flag=value`. This PR fixes this comment which appears as a field description when using `kubectl explain`.

Fixes #112

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/bootstrap/api/v1alpha1/rke2config_types.go
+++ b/bootstrap/api/v1alpha1/rke2config_types.go
@@ -335,7 +335,7 @@ type ComponentConfig struct {
 	//+optional
 	ExtraEnv map[string]string `json:"extraEnv,omitempty"`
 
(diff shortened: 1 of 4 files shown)


alexander-demicev (approved):
I removed `kind/documentation` field because our CI check accepts only one. We can revisit this approach.
Otherwise all looks great

PR #133, opened 2023-05-05
Improve e2e tests

lrascao:

`make test-e2e` did not work out of the box on my Debian 11, control plane booted up (with a noticeable delay) but worker node got perma stuck in `Provisioning`. After a bit of digging found `kubelet` was (silently due to `alsologtostderr=false` parameter) failing with an `inotify` error due to too many open files.
This patch intends to:
  * Check low default values on test start, alert user and suggest remediation
  * Decrease control plane bootstrap time by not failing status update on reconciliation when nodes are not present/not ready, failing prevents reconciliation requeing. In all test runs control machine is ready and with IP address but controller only sees this minutes after.

**Checklist**:

- [X] adds or updates e2e tests


controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -339,11 +339,13 @@ func (r *RKE2ControlPlaneReconciler) updateStatus(ctx context.Context, rcp *cont
 	}
 
 	if len(ownedMachines) == 0 {
-		return fmt.Errorf("no Control Plane Machines exist for RKE2ControlPlane %s/%s", rcp.Namespace, rcp.Name)
+		r.Log.Info(fmt.Sprintf("no Control Plane Machines exist for RKE2ControlPlane %s/%s", rcp.Namespace, rcp.Name))
```
alexander-demicev, line 342: Please use `logger` defined at the beginning of this method. We need to clean up our logging, it a bit messy right now. It will already include name and namespace of the object.

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -339,11 +339,13 @@ func (r *RKE2ControlPlaneReconciler) updateStatus(ctx context.Context, rcp *cont
...
+		return nil
 	}
 
 	if len(readyMachines) == 0 {
-		return fmt.Errorf("no Control Plane Machines are ready for RKE2ControlPlane %s/%s", rcp.Namespace, rcp.Name)
+		r.Log.Info(fmt.Sprintf("no Control Plane Machines are ready for RKE2ControlPlane %s/%s", rcp.Namespace, rcp.Name))
+		return nil
```
belgaied2, line 348: This changed, combined with the previous change, will cause a wrongly set `rcp.Status.Ready`.
Indeed, when you continue the logic without triggering a `Requeue`, you will find this:

```go
if len(readyMachines) == len(ownedMachines) {
	rcp.Status.Ready = true
}
```
If both `ownedMachines` and `readyMachines` collections are empty, the if condition will be true, and the `rcp.Status.Ready` will be `true` : this is not the desired behaviour. We need to keep requeuing whenever `ownedMachines` or `readyMachines` is empty.

belgaied2 (changes requested):
Removing requeuing in the RKE2ControlPlane `updateStatus` method is not the desired behaviour.

controlplane/internal/controllers/rke2controlplane_controller.go
lrascao: updated in 4a78a8a3204fbe45a8a0320f44f6f9861b908c8c

controlplane/internal/controllers/rke2controlplane_controller.go
lrascao: If either one of `ownedMachines` or `readyMachines` is empty we still return, what this patch does is return with a `nil` error, so we won't hit the if condition you mention. Requeing was not happening before this patch, if either one of collection is empty we'd return an error from `updateStatus` that gets assigned to `reterr`:
```
		if err := r.updateStatus(ctx, rcp, cluster); err != nil {
			var connFailure *rke2.RemoteClusterConnectionError
			if errors.As(err, &connFailure) {
				logger.Info("Could not connect to workload cluster to fetch status", "err", err.Error())
			} else {
				logger.Error(err, "Failed to update RKE2ControlPlane Status")
				reterr = kerrors.NewAggregate([]error{reterr, err})
			}
		}
```

Due to `reterr != nil` we won't requeue (with the default time) thus leading to high times of control plane readiness:

```
		if reterr == nil && !res.Requeue && res.RequeueAfter <= 0 && rcp.ObjectMeta.DeletionTimestamp.IsZero() {
			if !rcp.Status.Ready {
				res = ctrl.Result{RequeueAfter: DefaultRequeueTime}
			}
		}
```

belgaied2: You are right in terms of behaviour, I didn't read well the return stuff.
Now, I am wondering in terms of code readability and meaning. Is it better to:
- return an error and handle it in the `defer` func after the` r.updateStatus` call
- return nil and keep the `defer `func as it is

And I do lean into the second one (your approach) because not having ready machines is not an actual failure, it just needs a requeue.

belgaied2:
@lrascao After my last comment, we are good and ready to merge after you fix the non-passing checks. You have a couple of minor linting issues and you need to make sure all commits are signed off.
Thank you again for your contribution, this is very appreciated!

lrascao:
@belgaied2 ready for another test round (rebased, squashed and signed)

alexander-demicev (approved):

PR #153, opened 2023-06-30
fix: incorrectly defaulting advertise address

richardcase:

With the recent change to introduce new registration methods we started to default the `advertiseAddress` to the "registration address" if the registration method of `address` was used. This caused the kube-api server to be startedw tith that address. If you then used a VIP/LB solution like kube-vip or metalb that runs within cluster it caused pods to not start start as there is a chicken and egg scenario.

This change removes that defaulting and also adds a sample that uses kube-vip.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


tmmorin:
This looks very good!

We'll test this out as soon as this ships.

richardcase:
NOTE: merged directly instead of waiting for reviews so that a release could be done asap.

furkatgofurov7 (commented):
LGTM

PR #180, opened 2023-10-17
Butane ignition v3

alknopfler:
Add support for ignition v3_3 using Butane instead of CLC. 
This implies extend the ignition interface to use butane from coreos

Fixes #

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -154,17 +154,33 @@ type RKE2AgentConfig struct {
...
+	// ButaneConfig contains butane specific configuration.
 	// +optional
-	Config string `json:"config,omitempty"`
+	ButaneConfig *ButaneConfig `json:"butaneConfig,omitempty"`
 
-	// Strict controls if Config should be strictly parsed. If so, warnings are treated as errors.
+	// Ignition version, Default to clc
+	// +kubebuilder:validation:Enum=butane
+	// +optional
+	Transpiler string `json:"transpiler,omitempty"`
```
hardys, line 169: Perhaps I'm misunderstanding, but today doesn't `additionalUserData` accept ignition JSON directly?  AFAICS the CLC usage is just internal templating, so we could just remove it and avoid adding this new dependency on Butane? 

@alexander-demicev perhaps you can confirm?

alexander-demicev: `AdditionalUserData` accepts CLC config that is later transformed into Ignition. I think we can avoid changing anything in the API objects. If we want to drop support for Ignition v2, let's leave the API as it is, remove the CLC package, and add Butane to replace it on the backend.
This way there will be no breaking changes on the API object itself, only the format of `AdditionalUserData.Config` will change(butane instead of clc) and on the backend butane will be used for templating.

alknopfler: we need ignition v3 for sure because Sle Micro only support v3

alexander-demicev: sure, so we can drop v2 entirely on the backend by removing CLC and using Butane, and on the API side no changes are needed

alknopfler: done!

alknopfler:
@alexander-demicev could you please take a look the changes to keep the api as is, adding the butane backend???

alknopfler:
changed to fcos 1.4 which is equivalent to ignition v3.3 (required 3.3 for slemicro 5.5)

alknopfler:
tests locally passed and cluster has been provisioned with the ignition from template adding some custom additionalConfig about root password override.

bootstrap/internal/ignition/butane/butane.go
```diff
@@ -203,66 +209,61 @@ func renderCLC(input *cloudinit.BaseUserData) ([]byte, error) {
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "rendering CLC configuration")
+		return ignitionTypes.Config{}, "", fmt.Errorf("error converting to Ignition: %v", err)
 	}
-
-	userData, _, err := buildIgnitionConfig(clcBytes, additionalConfig)
+	if (len(reports.Entries) > 0 && strict) || reports.IsFatal() {
+		return ignitionTypes.Config{}, "", fmt.Errorf("error converting to Ignition: %v", reports.String())
+	}
+	log.Println(string(ignBytes))
```
alexander-demicev, line 224: should this be removed?

alexander-demicev (commented):
great work, thanks! Can you please squash commits in this branch?

alexander-demicev (approved):

Danil-Grigorev (approved):

PR #236, opened 2024-01-12
fix: rke2 registration token generated twice

### LINKED ISSUE #233: RKE2 registration token inconsistency between 1st CP and following nodes
**What happened:**

When using CAPRKE2 with CAPO there is an issue where the registration token on the 1st CP machine differs to that of the following CP machine or the worker machines. This results in a failure to join the new nodes to cluster with a 401 not authorised

Expectation is that the nodes are able to join the cluster with the correct token.

**How to reproduce it:**

See [sylva downstream issue](https://gitlab.com/sylva-projects/sylva-core/-/issues/832) for more detail.

**Environment:**

- rke provider version:

salasberryfin:

The RKE2 Config controller is creating the registration token twice (intermittently) and updating the secret that contains the token so different nodes are trying to get registered with different tokens and failing to do so.

We must ensure [this secret](https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/c69285302e9d3733559f65a2ef8031935e0144e4/bootstrap/internal/controllers/rke2config_controller.go#L337) is created once and its value remains unedited so that nodes can be registered to the cluster.

This initial proposal suggests detecting if the secret exists and skip token creation if it does.

Fixes #233 

We should add test cases that cover the logic of this controller: #https://github.com/rancher-sandbox/cluster-api-provider-rke2/issues/235 tracks this follow-up issue.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/bootstrap/internal/controllers/rke2config_controller.go
+++ b/bootstrap/internal/controllers/rke2config_controller.go
@@ -334,14 +334,28 @@ func (r *RKE2ConfigReconciler) handleClusterNotInitialized(ctx context.Context, 
 
 	conditions.MarkTrue(scope.Config, bootstrapv1.CertificatesAvailableCondition)
 
-	token, err := r.generateAndStoreToken(ctx, scope)
+	// RKE2 server token must only be generated once, so all nodes join the cluster with the same registration token.
+	var token string
+
+	tokenName := bsutil.TokenName(scope.Cluster.Name)
+	token, err := r.generateAndStoreToken(ctx, scope, tokenName)
+
 	if err != nil {
-		scope.Logger.Error(err, "unable to generate and store an RKE2 server token")
+		if !apierrors.IsAlreadyExists(err) {
+			scope.Logger.Error(err, "unable to generate and store an RKE2 server token")
 
-		return ctrl.Result{}, err
+			return ctrl.Result{}, err
+		}
+
+		token, err = r.getRegistrationTokenFromSecretValue(ctx, tokenName, scope.Cluster.Namespace)
+		if err != nil {
+			scope.Logger.Error(err, "unable to retrieve an RKE2 server token from existing secret")
+
+			return ctrl.Result{}, err
+		}
+	} else {
+		scope.Logger.Info("RKE2 server token generated and stored in Secret!")
 	}
-
-	scope.Logger.Info("RKE2 server token generated and stored in Secret!")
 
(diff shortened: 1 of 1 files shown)


bootstrap/internal/controllers/rke2config_controller.go
```diff
@@ -334,14 +334,28 @@ func (r *RKE2ConfigReconciler) handleClusterNotInitialized(ctx context.Context,
...
+	// RKE2 server token must only be generated once, so all nodes join the cluster with the same registration token.
+	var token string
 
-		return ctrl.Result{}, err
+	tokenSecret := &corev1.Secret{}
+	secretKey := types.NamespacedName{
+		Namespace: scope.Config.Namespace,
+		Name:      bsutil.TokenName(scope.Cluster.Name),
 	}
+	err := r.Client.Get(ctx, secretKey, tokenSecret)
```
tmmorin, line 345: I'm under the impression that this code might be subject to a race condition ; for instance in the case where we would run multiple replicas of the controller (both controller could fail the get in the same window of time before any reaches the create, and then both would do the create).

I think that the right pattern here would be to:
- try to create
- if this fails because the object already exists, then read it's content

Maybe one possibility is to make this update step of createOrUpdateSecretFromObject optional:
https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/main/bootstrap/internal/controllers/rke2config_controller.go#L827-L829

and call createOrUpdateSecretFromObject from generateAndStoreToken with this update step disabled

salasberryfin: Thanks for the feedback @tmmorin. My initial idea was to create a separate `createSecretFromObject` and call this function to generate the token, so that the update functionality is not altered and we don't have to deal with the value being updated, but decided to opt for this simpler solution. Didn't think about the corner case you described. I'll review this and take the other approach.

salasberryfin:
PR is now updated based on review feedback. Current logic will:
- Try to create secret and store registration token.
- If secret already exists, retrieve registration token value.

bootstrap/internal/controllers/rke2config_controller.go
tmmorin: Thanks for solving this problem.

Danil-Grigorev (approved):
lgtm

richardcase (approved):
Tested this locally and i didn't see the same duplicate token creation.

alexander-demicev (approved):

PR #251, opened 2024-01-24
fix: Handling of registries config to work as rke2

### LINKED ISSUE #253: Feature parity of rke2 registry configuration
https://docs.rke2.io/install/containerd_registry_configuration < the provider should support the same functionality when configuring registries like here where you can configure it either with user/password, tls or a combo of both. 
 
Current implementation doesn't support either or.

ekarlso:

Fixes #253 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/rke2/registries.go
+++ b/pkg/rke2/registries.go
@@ -17,6 +17,7 @@ package rke2
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
 	bootstrapv1 "github.com/rancher-sandbox/cluster-api-provider-rke2/bootstrap/api/v1beta1"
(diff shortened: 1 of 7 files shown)


pkg/rke2/registries.go
```diff
@@ -45,85 +47,96 @@ func GenerateRegistries(rke2ConfigRegistry RegistryScope) (*Registry, []bootstra
...
-			},
-			&tlsSecret,
-		)
-		if err != nil {
-			rke2ConfigRegistry.Logger.Error(err, "TLS Config Secret for the registry was not found!")
-
-			return &Registry{}, []bootstrapv1.File{}, err
+		registryConfig := RegistryConfig{}
+
+		if regConfig.TLS == (bootstrapv1.TLSConfig{}) && regConfig.TLS != (bootstrapv1.TLSConfig{}) {
```
richardcase, line 52: Its best to put this type of check into the "validating webhook" so that its validated when the user tries to apply the manifest.

The check would need to be in a coupld of places:

- [ValidateCreate](https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/e828ffa7bd9c300b7d7ced430b64e6bcb7c94d9f/bootstrap/api/v1beta1/rke2config_webhook.go#L67)
- [ValidateUpdate](https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/e828ffa7bd9c300b7d7ced430b64e6bcb7c94d9f/bootstrap/api/v1beta1/rke2config_webhook.go#L82)

pkg/rke2/registries.go
```diff
@@ -45,85 +47,96 @@ func GenerateRegistries(rke2ConfigRegistry RegistryScope) (*Registry, []bootstra
...
+			err := rke2ConfigRegistry.Client.Get(
+				rke2ConfigRegistry.Ctx,
+				types.NamespacedName{
+					Name:      regConfig.TLS.TLSConfigSecret.Name,
+					Namespace: regConfig.TLS.TLSConfigSecret.Namespace,
+				},
+				&tlsSecret,
+			)
+			if err != nil {
+				rke2ConfigRegistry.Logger.Error(err, "TLS Config Secret for the registry was not found!")
```
salasberryfin, line 68: Perhaps we can check for a specific `NotFound` error here, so that we have clearer logging based on the type of error:
```
if apierrors.IsNotFound(err) {
    rke2ConfigRegistry.Logger.Error(err, "TLS Config Secret for the registry was not found!")
}
```
And a different log message for a failed `Get` request. What do you think?

ekarlso: I did this instead 
```
			if err != nil {
				if apierrors.IsNotFound(err) {
					rke2ConfigRegistry.Logger.Error(err, "TLS Config Secret for the registry was not found!")
				} else {
					rke2ConfigRegistry.Logger.Error(err, "Error fetching TLS Config Secret")
				}

				return &Registry{}, []bootstrapv1.File{}, err
			}```

richardcase (approved):

PR #265, opened 2024-02-14
🐛 Reconcile etcd members on control plane scale down

### LINKED ISSUE #263: Control plane nodes scale down causes etcd to loose quorum and do not restore
**What happened:**
[A clear and concise description of what the bug is.]
After initial provisioning, scaling down control plane replicas to 1 causes API server to stop responding and never restore.

etcd logs show this:
```
{"level":"warn","ts":"2024-02-12T12:18:56.818575Z","caller":"etcdserver/server.go:2085","msg":"failed to publish local member to cluster through raft","local-member-id":"56c279675ce0ee85","local-member-attributes":"{Name:caprke2-e2e-60mj1n-control-plane-xdffd-d9da8b76 ClientURLs:[https://172.18.0.10:2379]}","request-path":"/0/members/56c279675ce0ee85/attributes","publish-timeout":"15s","error":"etcdserver: request timed out"}
{"level":"warn","ts":"2024-02-12T12:18:56.904853Z","caller":"rafthttp/probing_status.go:68","msg":"prober detected unhealthy status","round-tripper-name":"ROUND_TRIPPER_RAFT_MESSAGE","remote-peer-id":"5320353a7d98bdee","rtt":"0s","error":"dial tcp 172.18.0.4:2380: connect: no route to host"}
{"level":"warn","ts":"2024-02-12T12:18:56.904964Z","caller":"rafthttp/probing_status.go:68","msg":"prober detected unhealthy status","round-tripper-name":"ROUND_TRIPPER_SNAPSHOT","remote-peer-id":"5320353a7d98bdee","rtt":"0s","error":"dial tcp 172.18.0.4:2380: connect: no route to host"}
{"level":"warn","ts":"2024-02-12T12:18:56.9283Z","caller":"rafthttp/probing_status.go:68","msg":"prober detected unhealthy status","round-tripper-name":"ROUND_TRIPPER_RAFT_MESSAGE","remote-peer-id":"ef183bce98bbb166","rtt":"0s","error":"dial tcp 172.18.0.9:2380: connect: no route to host"}
{"level":"warn","ts":"2024-02-12T12:18:56.930688Z","caller":"rafthttp/probing_status.go:68","msg":"prober detected unhealthy status","round-tripper-name":"ROUND_TRIPPER_SNAPSHOT","remote-peer-id":"ef183bce98bbb166","rtt":"0s","error":"dial tcp 172.18.0.9:2380: connect: no route to host"}
```

`caprke2-e2e-60mj1n-control-plane-xdffd-d9da8b76` is a node which was recently removed, yet the server infinitely tries to access it.

Cluster control plane node scaling to not affect API server health.

**How to reproduce it:**
- Create an RKE2 cluster with 3 control plane replicas.
- Wait for cluster to become available.
- Scale down CP to 1.
- 

**Environment:**

- rke provider version: `rke2 version v1.28.1+rke2r1`

Danil-Grigorev:
kind/bug

This change allows to establish connectivity to child cluster etcd members, and manage membership during cluster scaling. Specifically this is required when a cluster etcd leader is removed by scaling down procedure, as this causes the cluster API server to become unavailable and never come back online.

Therefore the etcd leader needs to be moved just before requesting node deletion, and etcd membership has to be adjusted as well.

RKE2 follows a [different](https://docs.rke2.io/security/certificates) certificate management model opposed to CAPI, therefore we can’t provide CAPI certificates on node bootstrap, and instead have to fetch the generated certificate by the rke2 server during bootstrapping.

New clusters will use regular CAPI certificate management [model](https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates#using-custom-certificates), where the certificates are provided to the RKE2 agent on the initialization and are generated if missing by cabprke2. Therefore, for the time being there will be 2 co-existing cluster configurations, which will be reduced to the upstream one by performing certificate rotation in the future.

Existing clusters will not be required to perform scale up in order for the fix to take effect. Regular upgrade will do this, even with `MaxSurge=0` setting. For them, if both local etcd secret, and child bootstrapped etcd secrets are missing, no etcd operations will be performed. First upgraded node, however, will be the “scale up” to populate the child cluster secret. This way it is guarantied that every cluster will eventually get the fix.

Fixes #263 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


Danil-Grigorev:
Current implementation is passing e2e tests with testing a single node upgrade and scaling to 1 scenarios. This will allow to preserve existing clusters, but in order to apply the fix, the cluster control plane replicas will have to be scaled up one node at some moment. Will explore different approach with passing generated certificates to the rke2 server. This may be closer to the upstream, but will require all existing clusters to be re-created, or have some etcd migration mechanism (like rke2 certificate rotation to manually supplied, if such thing is supported).
```
Ran 2 of 2 Specs in 6350.292 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

Danil-Grigorev:
@richardcase Presubmit e2e job passed without issues here :) And under 30 minutes.

alexander-demicev (dismissed):
thanks a lot for taking care of this problem

salasberryfin (dismissed):

test/e2e/e2e_test.go
```diff
@@ -168,46 +175,37 @@ var _ = Describe("Workload cluster creation", func() {
...
-			// }, result)
-
-			// WaitForControlPlaneToBeReady(ctx, WaitForControlPlaneToBeReadyInput{
-			// 	Getter:       bootstrapClusterProxy.GetClient(),
-			// 	ControlPlane: result.ControlPlane,
-			// }, e2eConfig.GetIntervals(specName, "wait-control-plane")...)
-
 			By("Scaling worker nodes to 1")
 
 			ApplyClusterTemplateAndWait(ctx, ApplyClusterTemplateAndWaitInput{
```
rumycoding, line 211: TODO in line 222 needs to be fixed

Danil-Grigorev:
e2e are failing after rebase and the output is very cryptic. I’m also seeing a bug in the RKE2 code (non-changed) related to `PodFailedReason` condition - once it is set, it is no longer updated, causing precondition checks to fail.

salasberryfin (approved):

pkg/rke2/workload_cluster.go
```diff
@@ -55,30 +68,84 @@ type WorkloadCluster interface {
...
 		Client:           cl,
 		Nodes:            map[string]*corev1.Node{},
 		nodePatchHelpers: map[string]*patch.Helper{},
 	}
+
+	restConfig = rest.CopyConfig(restConfig)
+	restConfig.Timeout = remoteEtcdTimeout
+
+	// Retrieves the etcd CA key Pair
+	keyPair, err := m.getEtcdCAKeyPair(ctx, clusterKey)
```
alexander-demicev, line 106:
```suggestion
	etcdKeyPair, err := m.getEtcdCAKeyPair(ctx, clusterKey)
```

alexander-demicev (approved):
thanks for working on this

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -518,6 +549,76 @@ func (r *RKE2ControlPlaneReconciler) reconcileNormal(
...
+		return nil
+	}
+
+	workloadCluster, err := r.GetWorkloadCluster(ctx, controlPlane)
+	if err != nil {
+		// Failing at connecting to the workload cluster can mean workload cluster is unhealthy for a variety of reasons such as etcd quorum loss.
+		return errors.Wrap(err, "cannot get remote client to workload cluster")
+	}
+
+	parsedVersion, err := semver.ParseTolerant(controlPlane.RCP.Spec.AgentConfig.Version)
```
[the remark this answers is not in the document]
salasberryfin: I think, during the time we need to support both, we can just set a priority on one of the fields and ignore the other (unless the priority one is missing). This would probably make it simpler to then propagate the field to where it is used and won't require much change in code like the above. Just thinking out loud after seeing this, more of a comment than a review.

Danil-Grigorev:
Tests are green again, there were some issues with the code, some problems are however in CAPI framework. It appears multiple machine sets are created per machine deployments in upgrade scenario. Testing code is unable to distinguish between those.

Kube-vip is not helpful as the pods are getting evicted due to MemoryPressure on the child cluster nodes, and the default set of tolerations are for some reason ignored there. Sometimes tests may flake, because with no load balancing solution may cause RKE2 agent to connect on a non-existing (recently removed) node. Other issue I observed with kube-vip, is that leader election is never able to release a lock from a dead pod on a node where the etcd instance is offline. This is likely a client-go issue.

That being said, the PR is ready to be merged, as the functionality is consistent with description.

richardcase (approved):
Looks ok to me.

@Danil-Grigorev - we should also follow-up on the `Kube-vip is not helpful` comment.

furkatgofurov7 (approved):
Thanks, LGTM

PR #300, opened 2024-04-18
Let the rke2-install.sh fail for pipe errors and undefined variables

### LINKED ISSUE #276: Installation with ignition can fail with a misleading error
**What happened:**
When deploying on SLEMicro with ignition I hit an error like this coming from `rke2-install.service`

```
+ semanage fcontext -a -t systemd_unit_file_t /usr/lib/systemd/system/rke2-server.service
ValueError: File context for /usr/lib/systemd/system/rke2-server.service already defined
```

This is misleading as ` /usr/lib/systemd/system/rke2-server.service` doesn't actually exist - initially I suspected an selinux issue, but the problem happens earlier, the script looks like:

```
cat /etc/rke2-install.sh 
#!/bin/bash
set -e

curl -sfL https://get.rke2.io/ | INSTALL_RKE2_VERSION=v1.25.3+rke2r1 sh -s - server
semanage fcontext -a -t systemd_unit_file_t  /usr/lib/systemd/system/rke2-server.service
setenforce 0
systemctl enable rke2-server.service
systemctl start rke2-server.service
restorecon /etc/systemd/system/rke2-server.service
mkdir -p /run/cluster-api && echo success > /run/cluster-api/bootstrap-success.complete
setenforce 1
```

Because we don't set `pipefail` if anything causes the curl to fail, we silently ignore it and see the semanage error previously mentioned instead.

If the curl download fails, we should surface that error so the root cause can be addressed (in my case it was a network configuration issue)

Probably we should either set `pipefail` in the script, or split the `curl` and running the script into two steps

**How to reproduce it:**
Deploy in any environment where the connection to `https://get.rke2.io/` is somehow broken so the `curl` fails

**Environment:**

- rke provider version: 0.2.4
- OS (e.g. from `/etc/os-release`): SLEMicro 5.5

mchiappero:
The rke2-install.sh script sets the exit on error flag ('set -e'), however the script might fail to exit when a command in a pipe fails. Add the '-u' and '-o pipefail' flags to better identify errors within the install script.

Fixes #276

### DIFF
--- a/bootstrap/internal/ignition/butane/butane.go
+++ b/bootstrap/internal/ignition/butane/butane.go
@@ -108,7 +108,7 @@ storage:
       contents:
         inline: |
           #!/bin/bash
-          set -e
+          set -euo pipefail
           {{ range .PreRKE2Commands }}
           {{ . | Indent 10 }}
           {{- end }}


richardcase (approved):

richardcase:
@mchiappero - normally we require commits to be signed.

PR #302, opened 2024-04-19
Check rke2-serving secret to determine controlPlane.Status.Initialized

anmazzotti:
kind/bug

This should fix the behavior around setting the `controlPlane.Status.Initialized` flag. 
As per [documentation](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/control-plane#required-status-fields), this flag is:
```text
a boolean field that is true when the target cluster has completed initialization such that at least once, the target's control plane has been contactable.
```

The implementation is in line with Kubeadm, but instead of checking if the `kubeadm-config` ConfigMap was uploaded on the cluster, we check if the `rke2-serving` secret was.
This should be equivalent and as well a good enough marker that the control plane was indeed initialized.

**Which issue(s) this PR fixes**:
This should fix the RKE2 control plane provider deadlocking when disabling the `cloudController` kubernetes component.

```yaml
  serverConfig:
    disableComponents:
      kubernetesComponents:
        - cloudController
```

Disabling this component will forbid RKE2 from applying the `node.Spec.ProviderID` automatically. CAPI Infrastructure providers should manage that instead, but by contract they need the `ControlPlaneInitialized` condition to be true before attempting. This is the cause of the deadlock.

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controlplane/internal/controllers/rke2controlplane_controller.go
+++ b/controlplane/internal/controllers/rke2controlplane_controller.go
@@ -369,7 +369,16 @@ func (r *RKE2ControlPlaneReconciler) updateStatus(ctx context.Context, rcp *cont
 	rcp.Status.ReadyReplicas = int32(len(readyMachines))
 	rcp.Status.UnavailableReplicas = replicas - rcp.Status.ReadyReplicas
 
-	if rcp.Status.ReadyReplicas > 0 {
+	workloadCluster, err := r.getWorkloadCluster(ctx, util.ObjectKey(cluster))
+	if err != nil {
+		logger.Error(err, "Failed to get remote client for workload cluster", "cluster key", util.ObjectKey(cluster))
+
+		return fmt.Errorf("getting workload cluster: %w", err)
+	}
+
+	status := workloadCluster.ClusterStatus(ctx)
+
+	if status.HasRKE2ServingSecret {
 		rcp.Status.Initialized = true
 	}
 
@@ -790,11 +799,11 @@ func (r *RKE2ControlPlaneReconciler) reconcileControlPlaneConditions(
 		return ctrl.Result{}, nil
 	}
 
-	workloadCluster, err := r.managementCluster.GetWorkloadCluster(ctx, util.ObjectKey(controlPlane.Cluster))
+	workloadCluster, err := r.getWorkloadCluster(ctx, util.ObjectKey(controlPlane.Cluster))
 	if err != nil {
-		logger.Error(err, "Unable to get Workload cluster")
+		logger.Error(err, "Failed to get remote client for workload cluster", "cluster key", util.ObjectKey(controlPlane.Cluster))
 
-		return ctrl.Result{}, errors.Wrap(err, "cannot get remote client to workload cluster")
+		return ctrl.Result{}, fmt.Errorf("getting workload cluster: %w", err)
 	}
 
 	defer func() {
@@ -842,11 +851,11 @@ func (r *RKE2ControlPlaneReconciler) upgradeControlPlane(
 		return ctrl.Result{}, nil
 	}
 
-	workloadCluster, err := r.managementCluster.GetWorkloadCluster(ctx, util.ObjectKey(cluster))
+	workloadCluster, err := r.getWorkloadCluster(ctx, util.ObjectKey(cluster))
 	if err != nil {
-		logger.Error(err, "failed to get remote client for workload cluster", "cluster key", util.ObjectKey(cluster))
+		logger.Error(err, "Failed to get remote client for workload cluster", "cluster key", util.ObjectKey(cluster))
 
-		return ctrl.Result{}, err
+		return ctrl.Result{}, fmt.Errorf("getting workload cluster: %w", err)
 	}
 
 	if err := workloadCluster.InitWorkload(ctx, controlPlane); err != nil {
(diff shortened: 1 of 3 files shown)


anmazzotti:
Locally built and tested, I can see it working with the `cloudController` disabled and ProviderID set as expected:

```
NAME                       CLUSTER   NODENAME                                 PROVIDERID                                                   PHASE     AGE     VERSION
rke2-control-plane-l9fk5   rke2      m-651ddd3f-d4f4-48dc-919c-118d3e16c2c9   elemental://default/m-651ddd3f-d4f4-48dc-919c-118d3e16c2c9   Running   5m51s   v1.28.7
rke2-md-0-hkx5q-xxqql      rke2      m-8cc9f136-6849-4d44-89c3-8c26fa29b337   elemental://default/m-8cc9f136-6849-4d44-89c3-8c26fa29b337   Running   5m52s   v1.28.7+rke2r1
```

anmazzotti:
Still looking at the failing test.
Seems to be a rollout issue, one new Machine was created and in `Provisioning` phase, but the underlying DockerMachine failed to apply the bootstrap.

```yaml
status:
  conditions:
  - lastTransitionTime: 2024-04-22T08:41:07Z
    message: 1 of 2 completed
    reason: BootstrapFailed
    severity: Warning
    status: 'False'
    type: Ready
  - lastTransitionTime: 2024-04-22T08:41:07Z
    message: Repeating bootstrap
    reason: BootstrapFailed
    severity: Warning
    status: 'False'
    type: BootstrapExecSucceeded
  - lastTransitionTime: 2024-04-22T08:26:07Z
    status: 'True'
    type: ContainerProvisioned
```

On the failing container itself I can see a lot of:

```
Apr 22 09:04:58.039538 caprke2-e2e-5ioxg7-md-0-rvxmc-2nj9n rke2[264]: time="2024-04-22T09:04:58Z" level=info msg="Waiting to retrieve agent configuration; server is not ready: failed to get CA certs: Get \"https://127.0.0.1:6444/cacerts\": read
```

pkg/rke2/workload_cluster.go
```diff
@@ -266,6 +270,16 @@ func (w *Workload) ClusterStatus() ClusterStatus {
...
 
+	// find the rke2-serving secret
+	key := ctrlclient.ObjectKey{
+		Name:      rke2ServingSecretKey,
+		Namespace: metav1.NamespaceSystem,
+	}
+	err := w.Client.Get(ctx, key, &corev1.Secret{})
+
+	// In case of error we do assume the control plane is not initialized yet.
+	status.HasRKE2ServingSecret = err == nil
```
Danil-Grigorev, line 281:
```suggestion
        if err != nil {
	    logger := log.FromContext(ctx)
	    logger.Error(err, "Could not get rke2 serving secret")
	}
	status.HasRKE2ServingSecret = err == nil
```

anmazzotti: This may be a bit too spammy. 
Also this is not an error, in most cases we are just waiting here for the control plane to initialize.

If we want to emit a log, then I'd say it should be Info level, not even warning, because this will happen all times.

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -369,7 +369,16 @@ func (r *RKE2ControlPlaneReconciler) updateStatus(ctx context.Context, rcp *cont
...
+	workloadCluster, err := r.getWorkloadCluster(ctx, util.ObjectKey(cluster))
+	if err != nil {
+		logger.Error(err, "Failed to get remote client for workload cluster", "cluster key", util.ObjectKey(cluster))
+
+		return fmt.Errorf("getting workload cluster: %w", err)
+	}
+
+	status := workloadCluster.ClusterStatus(ctx)
+
+	if status.HasRKE2ServingSecret {
```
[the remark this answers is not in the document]
anmazzotti: In that case the controlPlaneInitialized should already be true, so if the secret is accidentally deleted, this condition will not be set to false. 
The `||` operator you suggested would re-introduce the condition I was trying to fix.

pkg/rke2/workload_cluster.go
Danil-Grigorev:
```suggestion
        if ctrlclient.IgnoreNotFound(err) != nil {
	    logger := log.FromContext(ctx)
	    logger.Error(err, "Unable to get RKE2 serving secret")
	}
	status.HasRKE2ServingSecret = err == nil
```
Probably for not found errors it might be spamming, can also be extended with other frequent err situations, but any other kind after the management cluster client was established should be treated as error. Leaving on your consideration, just think that converting error to bool without context can be difficult to track, and occasional *my CP is unavailable* issues will be more frequent due to failure of child cluster API server.

anmazzotti: I think the most common case that we will encounter in here is connection issues. This is the case where we just defined a cluster. The control-plane Machine may take a minute maybe to be created, then we have to wait for the physical machine to be bootstrapped, all processes running, kubelet alive and available. 

The entire process will take minutes maybe, during all this time this will fail due to connection errors, however this is to be expected. 

I do see your point however and I agree this should be highlighted to the end user. 
In that case I would just emit a log at Info level. We don't know if this is an error, it's for the user to judge. They will experience that the provider can't connect to the control plane (and therefore continue with the machines rollout). Maybe the user failed to configure the network properly, maybe we just have to wait, we don't know.

Something like this should work:

```go
        if err != nil {
	    logger := log.FromContext(ctx)
	    logger.Info("Control Plane is not initialized yet: %s", err.Error())
	}
```

Danil-Grigorev (approved):

Issue #196, opened 2023-11-15
[EPIC] v1beta2 API Changes


richardcase:
This is to track the various API change's we'd like to make in v1alpha2:

- [x] Reconsider the use of kubebuilder defaults, should we remove them to be more GitOps friendly. See discussion: https://github.com/rancher-sandbox/cluster-api-provider-rke2/pull/188/files#r1393845114
- [x] Remove version from the **AgentConfig**

PR #306, opened 2024-04-22
Use spec.Version field as primarily source

alexander-demicev:

Use spec.Version field as primarily source of RKE2 version. Currently, it's just a placeholder for satisfying API contracts. When both spec.AgentConfig.Version and spec.Version are specified, spec.Version take precedence.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/bootstrap/internal/controllers/rke2config_controller.go
+++ b/bootstrap/internal/controllers/rke2config_controller.go
@@ -278,6 +278,14 @@ type Scope struct {
 	ControlPlane         *controlplanev1.RKE2ControlPlane
 }
 
+func (s *Scope) getDesiredVersion() string {
+	if s.Machine.Spec.Version != nil && bsutil.IsRKE2Version(*s.Machine.Spec.Version) {
+		return *s.Machine.Spec.Version
+	}
+
+	return s.Config.Spec.AgentConfig.Version
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *RKE2ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if r.RKE2InitLock == nil {
@@ -363,6 +371,7 @@ func (r *RKE2ConfigReconciler) handleClusterNotInitialized(ctx context.Context, 
 			AgentConfig:          scope.Config.Spec.AgentConfig,
 			Ctx:                  ctx,
 			Client:               r.Client,
+			Version:              scope.getDesiredVersion(),
 		})
 	if err != nil {
 		return ctrl.Result{}, err
@@ -417,7 +426,7 @@ func (r *RKE2ConfigReconciler) handleClusterNotInitialized(ctx context.Context, 
 			PreRKE2Commands:         scope.Config.Spec.PreRKE2Commands,
 			PostRKE2Commands:        scope.Config.Spec.PostRKE2Commands,
 			ConfigFile:              initConfigFile,
-			RKE2Version:             scope.Config.Spec.AgentConfig.Version,
+			RKE2Version:             scope.getDesiredVersion(),
 			WriteFiles:              files,
 			NTPServers:              ntpServers,
 			AdditionalCloudInit:     scope.Config.Spec.AgentConfig.AdditionalUserData.Config,
@@ -567,6 +576,7 @@ func (r *RKE2ConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 			AgentConfig:          scope.Config.Spec.AgentConfig,
 			Ctx:                  ctx,
 			Client:               r.Client,
+			Version:              scope.getDesiredVersion(),
 		},
 	)
 	if err != nil {
@@ -627,7 +637,7 @@ func (r *RKE2ConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 			PreRKE2Commands:     scope.Config.Spec.PreRKE2Commands,
 			PostRKE2Commands:    scope.Config.Spec.PostRKE2Commands,
 			ConfigFile:          initConfigFile,
-			RKE2Version:         scope.Config.Spec.AgentConfig.Version,
+			RKE2Version:         scope.getDesiredVersion(),
 			WriteFiles:          files,
 			NTPServers:          ntpServers,
 			AdditionalCloudInit: scope.Config.Spec.AgentConfig.AdditionalUserData.Config,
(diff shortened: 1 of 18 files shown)


controlplane/api/v1beta1/rke2controlplane_types.go
```diff
@@ -436,3 +436,21 @@ func (r *RKE2ControlPlane) GetConditions() clusterv1.Conditions {
 func (r *RKE2ControlPlane) SetConditions(conditions clusterv1.Conditions) {
 	r.Status.Conditions = conditions
 }
+
+// GetDesiredVersion returns the desired version of the RKE2ControlPlane using Spec.Version field as a default field.
```
alexander-demicev, line 440: I added a separate function to the RKE2 CP package in case we want to reuse it somewhere else like day 2 operations

bootstrap/internal/controllers/rke2config_controller.go
```diff
@@ -278,6 +278,14 @@ type Scope struct {
 	ControlPlane         *controlplanev1.RKE2ControlPlane
 }
 
+func (s *Scope) getDesiredVersion() string {
+	if s.Machine.Spec.Version != nil && bsutil.IsRKE2Version(*s.Machine.Spec.Version) {
+		return *s.Machine.Spec.Version
+	}
+
+	return s.Config.Spec.AgentConfig.Version
```
richardcase, line 286: I updated #196 to remind us to remove the `AgentConfig.Version` field in the next version bump

bootstrap/internal/controllers/rke2config_controller.go
```diff
@@ -278,6 +278,14 @@ type Scope struct {
 	ControlPlane         *controlplanev1.RKE2ControlPlane
 }
 
+func (s *Scope) getDesiredVersion() string {
+	if s.Machine.Spec.Version != nil && bsutil.IsRKE2Version(*s.Machine.Spec.Version) {
```
richardcase, line 282: If there is a value `s.Machine.Spec.Version` but it doesn't have the rke2 suffix then it will be ignored? Should we raise an error in this situation?

controlplane/api/v1beta1/rke2controlplane_types.go
```diff
@@ -47,9 +49,7 @@ type RKE2ControlPlaneSpec struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 
 	// Version defines the desired Kubernetes version.
-	// This is only a placeholder for now, and the RKE2ConfigSpec.AgentConfig.Version field should be used instead.
-	// In future iterations, this field overrides the RKE2 Version specificied in RKE2ConfigSpec.AgentConfig.Version
-	// which will be deprecated in newer versions of the API.
+	// This field takes precedence over RKE2ConfigSpec.AgentConfig.Version.
```
richardcase, line 52:
```suggestion
	// This field takes precedence over RKE2ConfigSpec.AgentConfig.Version (which is deprecated).
```

richardcase (commented):
Generally looks good to me. 

We should add an upgrade test based on how people are currently using the provider to make sure we don't break anything.

bootstrap/internal/controllers/rke2config_controller.go
Danil-Grigorev: I think we should get rid of the IsRKE2Version and add a validation on the CRD itself. It will simplify things

alexander-demicev: @richardcase Currently we create machines with plain k8s version by removing RKE2 suffix. This function ensures nothing breaks existing clusters, we need to read version from agentConfig. After users migrate to `spec.Version` field and rollout new machines, `machine.Spec.Version` will always be RKE2 version.

alexander-demicev: @Danil-Grigorev we can't add CRD validation for core CAPI resources, the version for workers will come from MD. It's possible to create a custom webhook but I don't know if want to go this route.

alexander-demicev: We can still add a validation to the RKE2 Control Plane object, I will update the PR

Danil-Grigorev: CP provider doesn't care what version workers have, these will not appear here

controlplane/api/v1beta1/rke2controlplane_types.go
salasberryfin: Would it be valid if `GetDesiredVersion` (cp) and `getDesiredVersion` (bootstrap) return different versions? If this case is possible.

salasberryfin (approved):
Looks good to me.

controlplane/api/v1beta1/rke2controlplane_types.go
Danil-Grigorev: Thats also a good place for validation

Danil-Grigorev (changes requested):
rke2CP.Spec.Version is still populated from AgentConfig on CP creation

alexander-demicev:
@Danil-Grigorev this should stay a valid option, I don't want to deprecate a feature for users without notice. We can make it only possible place to specify version in future releases.

bootstrap/internal/controllers/rke2config_controller.go
alexander-demicev: We pass value from machine.Spec.Version to the bootstrap script for all type of machines, it means that this value must be a valid RKE2 version

alexander-demicev: I added the validation in the CRD but we still need to keep checking of RKE2 version in these functions for compatibility with already created clusters.

alexander-demicev:
@richardcase The Current upgrade test covers transition from the old version field to new one and then also upgrades the cluster with it

Danil-Grigorev (commented):
> @Danil-Grigorev this should stay a valid option, I don't want to deprecate a feature for users without notice. We can make it only possible place to specify version in future releases.

From what I see in the code this is not an option but a mandatory, it seems only internal handling was changed in this PR, but not external interfaces. I think that we are not following CAPI pattern with versions, the RKE2 provider can’t have different versions for new CP nodes and new workers after a CP reached a updated version. CP version IMO should set both, and we shouldn’t get the version from the machine. It looks strange to me.

salasberryfin (approved):

controlplane/api/v1beta1/rke2controlplane_types.go
```diff
@@ -436,3 +437,21 @@ func (r *RKE2ControlPlane) GetConditions() clusterv1.Conditions {
 func (r *RKE2ControlPlane) SetConditions(conditions clusterv1.Conditions) {
 	r.Status.Conditions = conditions
 }
+
+// GetDesiredVersion returns the desired version of the RKE2ControlPlane using Spec.Version field as a default field.
+func (r *RKE2ControlPlane) GetDesiredVersion() string {
+	if r.Spec.Version != "" && isRKE2Version(r.Spec.Version) {
```
Danil-Grigorev, line 443: This does not need validation `isRKE2Version` now that it is ensured on CRD level.

PR #313, opened 2024-04-30
🐛 Add old AgentConfig version conversion in favor of a new field

### LINKED ISSUE #196: [EPIC] v1beta2 API Changes

Danil-Grigorev:
This ensures we remove old AgentConfig version without breaking changes by using conversion webhook and the fact that we ship new API version in the next release.

For users affected by the #315 this will require 2 step process:
1. Check that version defined in `MachineDeployment` matches rke2 releases: https://github.com/rancher/rke2/releases
2. Force re-rollout of all worker nodes to the version currently set in `MachineDeployment` or upgrade workers to the new version.

Additionally, as per issue, removing use of defaults in the new API.

Fixes #196, #315

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


bootstrap/api/v1beta1/rke2config_types.go
```diff
@@ -146,12 +146,6 @@ type RKE2AgentConfig struct {
 	//+optional
 	LoadBalancerPort int `json:"loadBalancerPort,omitempty"`
 
-	// Version specifies the rke2 version.
-	// This field will be deprecated in newer versions of the API and RKE2ControlPlaneSpec.Version will be used instead.
-	// +kubebuilder:validation:Pattern="v(\\d\\.\\d{2}\\.\\d)\\+rke2r\\d"
-	//+optional
-	Version string `json:"version,omitempty"`
```
alexander-demicev, line 153: We can leave the conversion of version for the RKE2ControlPlane but removing `Version` from RKE2Config for workers is a breaking change because version is managed from MachineDeployment object. I know the API is in alpha but we have users running it in production.

alexander-demicev (changes requested):

bootstrap/api/v1beta1/rke2config_types.go
alexander-demicev: In previous versions of CAPRKE2, we took the version from the `RKE2Config.AgentConfig.Version` to pass to the installation script https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/v0.2.7/bootstrap/internal/controllers/rke2config_controller.go#L745 for worker machines. MachineDeployment version field is placeholder field in all previous versions, which means that not everyone had this field configured properly(or configured at all, that's also a valid option). By removing `RKE2Config.AgentConfig.Version` we introduce a breaking change where users will lose data about current version of their worker machines.

Danil-Grigorev: For those users who don’t have the MachineDeployment `spec.version` field configured properly or at all it indeed will be a breaking change. Under CAPI semantics this is invalid, as CP versions and worker versions supposed to be independent from each other. The problem is however in case when the version was configured properly, but is different from the CP version. These users as per #315 actually don’t have the expected version on the new worker nodes provisioned after the diff.

With this change we can be sure that all existing and new users of `v1beta1` do have declared MD/CP version matching the components version provisioned for the node.

Danil-Grigorev:
All questions are resolved? @alexander-demicev

alexander-demicev:
@Danil-Grigorev yes, thanks

alexander-demicev (approved):

salasberryfin (approved):
LGMT. I'm thinking this change will need to be well documented and properly communicated to those that are already running and may be in the situation @alexander-demicev mentioned. How should we proceed with this?

alexander-demicev:
@salasberryfin This was already asked by another team, we will create an ADR before releasing this change

Danil-Grigorev:
@salasberryfin It was already done. Additionally, as I mentioned in the comments, this is NOT breaking for user properly utilizing machine deployment version field, which should be all of the users. It is far riskier to proceed to v1beta1 with a chance new users hitting the bug without knowing and having us to design a solution to fix it for them at a later day.

Danil-Grigorev:
Created https://github.com/rancher-sandbox/cluster-api-provider-rke2/issues/323, cc @alexander-demicev @salasberryfin

PR #319, opened 2024-05-10
Add airGappedChecksum field to check workload version before rollout

### LINKED ISSUE #297: in airGapped mode, ensure that the desired k8s version is used
We noticed [while developing Sylva](https://gitlab.com/sylva-projects/sylva-core/-/issues/1081) that, in airGapped mode (`spec.agentConfig.airGapped: true`), the combination of RKE2 bootstrap provider  and RK2 itself will result in using _whatever version of RKE2 is present in `/opt/rke2-artifacts`, even if this version does not match `spec.agentConfig.version`.

It creates a risk of unintendedly using an undesired version, and possibly unintendedly upgrading or downgrading, if the OS image is not selected carefully.

A safer behavior for air-gapped mode would be to abort RKE2 installation if the RKE2 version that can be found on disk does not match `spec.agentConfig.version`.

/cc @richardcase

Danil-Grigorev:
This change adds `AirGappedChecksum` field to agent config. By using this field, a user might declare expected sha256sum checksum for the air gapped manifests and prevent incomplete or different package installation on the host.

Fixes #297 
Depends on #313 for fuzzy test changes to make CI green

Internals of the `curl -sfL https://get.rke2.io` in the air-gapped mode already verify the contents of provided tar files based on the checksum from `sha256sum-<arch>.txt`. Therefore the `AirGappedChecksum` declares the expected checksum for this file only to connect API definition with the prepared host pre-bootstrap state.

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


bootstrap/api/v1beta1/rke2config_types.go
```diff
@@ -156,6 +156,11 @@ type RKE2AgentConfig struct {
 	// basically supposing that online container registries and RKE2 install scripts are not reachable.
 	AirGapped bool `json:"airGapped,omitempty"`
 
+	// AirGappedChecksum is a string value with a sha256sum checksum to compare with checksum
+	// of existing sha256sum-<arch>.txt file for packages already available on the machine
+	// before performing air-gapped installation.
+	AirGappedChecksum string `json:"airGappedChecksum,omitempty"`
```
alexander-demicev, line 162: can we make this field optional?

alexander-demicev (approved):
thanks, looks good just one small comment

salasberryfin:
LGTM pending fixing CI checks and resolving @alexander-demicev's comment.

bootstrap/api/v1beta1/rke2config_types.go
Danil-Grigorev: It was blocked on #313 for fuzzy testing, so I was only able to add generated manifests, make the field optional and backward value recovery after it got merged. This should be ready for review.

alexander-demicev (approved):

PR #335, opened 2024-05-21
Add ADR explaining removal of kubebuilder defaults

alexander-demicev:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


docs/adr/0002-deprecate-kubebuilder-defaults.md
```diff
@@ -0,0 +1,51 @@
...
+    type: "RollingUpdate"
+    rollingUpdate:
+      maxSurge: 1 # can be 0
+```
+
+- `spec.registrationMethod`:
+
+```yaml
+  registrationMethod:
+    type: "internal-first" # other supported values are "internal-only-ips", "external-only-ips", "address", "control-plane-endpoint"
```
Danil-Grigorev, line 50: Can we add that internal first is not recommended and may lead to issues with scaling? And that recommended approach is to use control-plane-endpoint and adjust load balancer configuration accordingly (if needed)?

alexander-demicev: done, can you take a look again?

docs/adr/0002-deprecate-kubebuilder-defaults.md
```diff
@@ -0,0 +1,52 @@
...
+<!-- markdown-link-check-disable-next-line -->
+- Status: accepted
+- Date: 2024-05-21
+- Authors: @alexander-demicev
+- Deciders: @Danil-Grigorev @furkatgofurov7 @salasberryfin @mjura @yiannistri
+
+## Context
+<!-- What is the context of the decision and what's the motivation -->
+
+Enforcing default values for fields in CRD definitions can cause problem with the GitOps approach. It can lead to a situation where defined resources in Git repository differs from the actual state of applied resources in the cluster. This can lead to confusion and unexpected behavior.
```
Danil-Grigorev, line 24:
```suggestion
Enforcing default values for fields in CRD definitions can cause problem with the GitOps approach. It can lead to a situation where defined resources in Git repository differs from the actual state of applied resources in the cluster. This can lead to unexpected behavior like state drift, unless is manually handled.
```

docs/adr/0002-deprecate-kubebuilder-defaults.md
```diff
@@ -0,0 +1,52 @@
...
+
+Users have to specify the following fields manually when creating RKE2ControlPlane resources:
+
+- `spec.rolloutStrategy`:
+
+```yaml
+  rolloutStrategy:
+    type: "RollingUpdate"
+    rollingUpdate:
+      maxSurge: 1 # can be 0
```
Danil-Grigorev, line 43: This one is defaulted to 1, so is not necessary to specify all the time.

Danil-Grigorev (approved):

docs/adr/0002-deprecate-kubebuilder-defaults.md
alexander-demicev: I decided to leave it here for awareness since we don't default this field in the object itselft

Danil-Grigorev (approved):

salasberryfin (commented):
This is just a question and not a blocker for merging the PR. Can this deprecation of default values cause an issue for users that have been using older versions of the provider? Let's say that a user provisions clusters via a GitOps tool (like Fleet) and they are missing some of the fields that no longer have a default value. Now they update CAPRKE2 to the new version that requires to set fields manually. Is it possible that this breaks their downstream cluster/s?

Danil-Grigorev:
@salasberryfin Defaults are only evaluated on the first apply, then the applied data is stored in etcd, so all existing manifests should already have user defined values for these fields.

salasberryfin:
Thanks @Danil-Grigorev.

salasberryfin (approved):

PR #357, opened 2024-07-05
Fix issue with registries required fields

alknopfler:

The registry TLS configuration requires the 3 values ca.crt, tls.crt, tls.key when only ca.crt is required (in case the registry does not need the client certs). With the current code, you're not able to include only the ca.crt. Also, there's no way to include the insecureSkipVerify flag to remove the rest of certs (we're skipping)

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/rke2/registries.go
+++ b/pkg/rke2/registries.go
@@ -29,6 +29,9 @@ const (
 	DefaultRKE2RegistriesLocation string = "/etc/rancher/rke2/registries.yaml"
 
 	registryCertsPath string = "/etc/rancher/rke2/tls"
+	cacert            string = "ca.crt"
+	tlskey            string = "tls.key"
+	tlscert           string = "tls.crt"
 )
 
 // GenerateRegistries generates the registries.yaml file and the corresponding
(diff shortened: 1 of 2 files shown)


pkg/rke2/registries.go
```diff
@@ -69,24 +69,38 @@ func GenerateRegistries(rke2ConfigRegistry RegistryScope) (*Registry, []bootstra
...
-				})
+				return &Registry{}, []bootstrapv1.File{}, err
 			}
 
-			registryConfig.TLS = &TLSConfig{
-				InsecureSkipVerify: regConfig.TLS.InsecureSkipVerify,
-				CAFile:             registryCertsPath + "/" + "ca.crt",
-				CertFile:           registryCertsPath + "/" + "tls.crt",
-				KeyFile:            registryCertsPath + "/" + "tls.key",
+			for _, secretEntry := range []string{"tls.crt", "tls.key", "ca.crt"} {
```
alexander-demicev, line 86: I'd like to recommend making "tls.crt", "tls.key", "ca.crt" constants so we don't retype the same strings if needed.

alexander-demicev (approved):
Thanks for the PR, only one small comment.

pkg/rke2/registries.go
alknopfler: changed!

alexander-demicev (approved):

Issue #401, opened 2024-08-12
cisProfile option broken with ignition format


hardys:
**What happened:**
Deploying with a configuration like this is currently broken due to formatting issues in the butane config:

```
    cisProfile: cis-1.23
    format: ignition
```

**What did you expect to happen:**
The `cisProfile` option should enable the `rke2-cis-script.sh` inside the rendered `rke2-install.sh` however due to stray tabs in the template it fails with an error like:

```
E0810 07:53:11.578031       1 controller.go:329] "Reconciler error" err="converting base config to Ignition: error converting to Ignition: yaml: line 712: found a tab character where an indentation space is expected" controller="rke2config" controllerGroup="bootstrap.cluster.x-k8s.io" 
```

**How to reproduce it:**

Deploy with `cisProfile` and ignition format enabled

**Environment:**
Only impacts ignition format deployments AFAICS, e.g Leap/SLEMicro

hardys:
I'm working on a fix for this so please assign the issue to me, thanks!

PR #301, opened 2024-04-18
Add "cis "to the cis-profile enum to support 1.29+

AshleyDumaine:
**What this PR does / why we need it**: For RKE2 version >=1.29, the cis profile flag needs to be "cis" (https://docs.rke2.io/security/hardening_guide#generic-cis-configuration), but this isn't currently in the CRD enums.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**: I'm not sure if a v1beta2 will be needed for this since this since this is a CRD change but no fields are being added/removed from the spec.

**Checklist**:

- [x] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #402, opened 2024-08-12
Fix cisProfile option with ignition format

hardys:
Currently there are some stray tabs which break the rendered output, replace these with spaces and add some unit test coverage.

kind/bug

Currently the cisProfile option is broken when selecting ignition format for Leap/SLEMicro deployments

Fixes: #401

**Checklist**:

- [ ] includes documentation
- [X] adds unit tests

### DIFF
--- a/bootstrap/internal/ignition/butane/butane.go
+++ b/bootstrap/internal/ignition/butane/butane.go
@@ -38,9 +38,8 @@ import (
 // The rke2-install.service unit is enabled and is executed only once during the boot process to run the /etc/rke2-install.sh script.
 // This script installs and deploys RKE2, and performs pre and post-installation commands.
 // The ntpd.service unit is enabled only if NTP servers are specified.
-// The second section defines storage files for the system. It creates a file at /etc/rke2-install.sh. If CISEnabled is set to true,
-// it runs an additional CIS script to enforce system security standards. If NTP servers are specified,
-// it creates an NTP configuration file at /etc/ntp.conf.
+// The second section defines storage files for the system. It creates a file at /etc/rke2-install.sh.
+// If NTP servers are specified, it creates an NTP configuration file at /etc/ntp.conf.
 const (
 	butaneTemplate = `
 variant: fcos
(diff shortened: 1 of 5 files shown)


hardys:
Review welcome but marking this as a draft until I've tested it e2e locally

hardys:
> @hardys let's rebase before re-running e2e tests on this one, thanks!

Will do, just want to complete some more local testing then will rebase and remove the draft status

Also looks like we'll need https://github.com/rancher/cluster-api-provider-rke2/pull/301 so good to see that may be getting revived

hardys:
After testing locally I found another issue:

```
	converting base config to Ignition: error converting to Ignition: warning at $.storage.files.1.mode, line 41 col 13: unreasonable mode would be reasonable if specified in octal; remember to add a leading zero
```

I think that's because the [mode constant here](https://github.com/rancher/cluster-api-provider-rke2/blob/main/pkg/consts/global_consts.go#L49) needs updating, will add to this PR and re-test

hardys:
I found some more issues which should now be resolved:

- The CIS prepration script was copies to /opt which isn't writable for Leap/SLEMicro, moved to /etc for both ignition/cloud-init cases so it's alongside the configuration files
- The script was running too early in the ignition flow, prior to the RKE2 installation, so the sysctl file was not yet present
- The script logic to find the sysctl file did not include the path used by the RKE2 installer, now added

hardys:
> The CIS prepration script was copies to /opt which isn't writable for Leap/SLEMicro, moved to /etc for both ignition/cloud-init cases so it's alongside the configuration files

The latest iteration takes a different approach and mounts the /opt subvolume as described in https://en.opensuse.org/Portal:MicroOS/Ignition#Mounts - I'm still not sure writing the file to /opt is ideal, but with this approach I can avoid changing any of the existing cloud-init related code.

hardys:
@alexander-demicev @furkatgofurov7 this is passing CI and now ready for review - please take a look when you get a moment, thanks!

alexander-demicev (approved):

furkatgofurov7 (approved):

PR #411, opened 2024-08-21
Overwrite files with format: ignition

### LINKED ISSUE #410: File overwrite fails with ignition format
**What happened:**

During deployment it's possible to define additional files with a configuration like the following, which when using cloud-init results in default behavior to [overwrite any existing files ](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#write-files)

```
---
apiVersion: bootstrap.cluster.x-k8s.io/v1alpha1
kind: RKE2ConfigTemplate
metadata:
  name: sample-cluster-workers
  namespace: default
spec:
  template:
    spec:
      files:
        - path: "/etc/motd"
          owner: "root:root"
          permissions: "0640"
          content: |
            hello 123
```

The behavior with `format: ignition` however is different and any attempt to overwrite an existing file will fail during the OS firstboot:

```
[   17.691712] worker-0 ignition[1180]: Ignition failed: failed to create files: failed to create files: error creating /sysroot/etc/motd: error creating file "/sysroot/etc/motd": A file exists there already and overwrite is false
```

The behavior between cloud-init and ignition should be consistent from the perspective of the user of the RKE2 provider API, even if the default behavior of the underlying tool may be different.

I think we can likely adapt the configuration during the translation from cloud-init to butane/ignition so that we set the overwrite flag to true.

**How to reproduce it:**
Deploy with a similar configuration described above and `format: ignition`

**Environment:**
I reproduced using Leap Micro 6.0 with CAPM3.

hardys:
This aligns with the [default cloud-init behavior](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#write-files) and enables e.g overwriting distro-default files via RKE2ConfigTemplate using the spec.template.spec.files API.

With this we can see the example mentioned in #410 now works as expected the empty `/etc/motd` is overwritten.

```
worker-0 login: root
Password: 
hello 123
```

kind/bug
Fixes #410

### DIFF
--- a/bootstrap/internal/ignition/butane/butane.go
+++ b/bootstrap/internal/ignition/butane/butane.go
@@ -95,6 +95,7 @@ storage:
       {{ if ne .Permissions "" -}}
       mode: {{ .Permissions }}
       {{ end -}}
+      overwrite: true
       contents:
         {{ if eq .Encoding "base64" -}}
         inline: !!binary |


hardys:
@alexander-demicev @furkatgofurov7 - this is passing CI and ready for review, when you get a moment please take a look - thanks!

furkatgofurov7 (approved):

furkatgofurov7:
Thanks @hardys for
the fix!

alexander-demicev (approved):

furkatgofurov7:
@hardys CI requires all commits to be signed before merging the PRs, can you please push it signed? Thanks.

hardys:
> @hardys CI requires all commits to be signed before merging the PRs, can you please push it signed? Thanks.

Now done, thanks!

furkatgofurov7 (approved):
Thank you!

alexander-demicev (approved):

hardys:
Sorry I misunderstood the signing requirement initially - commit should now be signed/verified https://github.com/rancher/cluster-api-provider-rke2/pull/411/commits

Issue #343, opened 2024-06-03
Improve usability of ControlPlane's registrationMethod with Cluster Classes


anmazzotti:
**Describe the solution you'd like:**
Due to validation of the `RKE2ControlPlane.spec.registrationMethod`, it is not currently possible to use `registrationMethod: "address"` in ClusterClasses without providing a valid `registrationAddress` too.

Even with patching this value afterwards, like:

```yaml
    - jsonPatches:
      - op: replace
        path: /spec/template/spec/registrationAddress
        valueFrom:
          variable: myRegistrationAddress
```

This will fail with:

```
The RKE2ControlPlane "rke2-control-plane" is invalid: spec.registrationAddress: Invalid value: "": registrationAddress must be supplied when using registration method 'address'
```

As the initial RKE2ControlPlane is created with empty `registrationAddress`, before patching.

One workaround is to configure the initial registrationMethod to `control-plane-endpoint` or any other default that does not require validation. 

Then I think one patch should replace both values ad the same time:

```yaml
    - jsonPatches:
      - op: replace
        path: /spec/template/spec/registrationMethod
        value: "address"
      - op: replace
        path: /spec/template/spec/registrationAddress
        valueFrom:
          variable: myRegistrationAddress
```

I did not test it, but it should technically work.
The drawback of this workaround however, is that initially the RKE2ControlPlane object is created with a different registrationMethod. This could lead to race condition issues where maybe nodes are joining using an unwanted registration method.

It should be fine on the initial cluster creation, assuming the first node ever will be the first control plane node, so there will be nothing to register on yet, and after that (or even before bootstrapping this node), the registrationMethod should be patched to the user defined value.

Could be more tricky however on a control plane rollout or during other upgrade scenarios.

**Why do you want this feature:**
I want to avoid having to configure a default `registrationMethod` for the real one to be patched later, if this could lead to problems.

PR #418, opened 2024-08-27
Make control plane endpoint default registration method

Danil-Grigorev:
Fixes: #343 

From the list of available registration methods, control plane one seems to be the most suitable for a default. It will not work for all cases, but for those which do, namely clusters without restrictions on accessing public API and address of CP endpoint, this should work out of the box as a quickstart method.

Users may decide to define a different registrationMethod, where one edit of the value is intentionally allowed.

Fixes #

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


alexander-demicev (dismissed):

controlplane/api/v1beta1/rke2controlplane_types.go
```diff
@@ -77,9 +77,9 @@ type RKE2ControlPlaneSpec struct {
 	NodeDrainTimeout *metav1.Duration `json:"nodeDrainTimeout,omitempty"`
 
 	// RegistrationMethod is the method to use for registering nodes into the RKE2 cluster.
-	// +kubebuilder:validation:Enum=internal-first;internal-only-ips;external-only-ips;address;control-plane-endpoint
+	// +kubebuilder:validation:Enum=internal-first;internal-only-ips;external-only-ips;address;control-plane-endpoint;""
 	// +optional
-	RegistrationMethod RegistrationMethod `json:"registrationMethod"`
+	RegistrationMethod RegistrationMethod `json:"registrationMethod,omitempty"`
```
furkatgofurov7, line 82: By PR description, the first thing that came to my mind was seeing something like: `+kubebuilder:default` for the `control-plane-endpoint` registration method. Is that a wrong assumption?

Danil-Grigorev: That would unfortunately break the ADR progression on this topic: https://github.com/rancher/cluster-api-provider-rke2/blob/main/docs/adr/0002-deprecate-kubebuilder-defaults.md#context - having a default value causes the field to be incompatible with gitops, since this field will cause a state drift by being defaulted by the API server

furkatgofurov7 (dismissed):

salasberryfin (dismissed):

furkatgofurov7:
This needs an update as discussed offline with @Danil-Grigorev, please hold off merging it manually.

Danil-Grigorev:
Upon checking crust-gather archive after test run, it seems the field is emptied, as expected. Just added e2e tests after all, leaving the rest of the logic unchanged. Should be good to merge now @furkatgofurov7 @salasberryfin 
```yaml

spec:
  agentConfig:
    additionalUserData: {}
    format: cloud-config
    nodeAnnotations:
      test: 'true'
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: DockerMachineTemplate
    name: caprke2-e2e-yqwygg-control-plane
  machineTemplate:
    infrastructureRef: {}
    metadata: {}
  manifestsConfigMapReference: {}
  nodeDrainTimeout: 30s
  privateRegistriesConfig: {}
  replicas: 1
  rolloutStrategy:
    rollingUpdate:
      maxSurge: 1
    type: RollingUpdate
  serverConfig:
    disableComponents:
      kubernetesComponents:
      - cloudController
    etcd:
      backupConfig: {}
  version: v1.28.12+rke2r1
status:
…
```

anmazzotti (approved):
If I read this correctly, an empty `registrationMethod` value is now equivalent to `control-plane-endpoint`, **but** it allows replacement, which should solve indeed my original issue.

I still wonder however if this is going to create race conditions, what if I don't want to use the `control-plane-endpoint` as registration method, but I want to directly use an address and input that address as variable in a cluster class? I don't see any example of this in the PR. 

I wonder if it's possible to put the control plane on "pause" until the `registrationMethod` is populated, but this seems to be an improvement.

alexander-demicev (approved):
LGTM, lets wait and see if @furkatgofurov7 has any other concerns

test/e2e/config/e2e_conf.yaml
```diff
@@ -44,6 +44,23 @@ providers:
           new: "--leader-elect=false"
         - old: --metrics-addr=127.0.0.1:8080
           new: --metrics-addr=:8080
+  - name: docker-updated
```
furkatgofurov7, line 47: I would be in favor of not duplicating this since it is a one liner less compared to the original cluster template we have in the repo, however, it looks like bigger of a change to achieve this (using jinja templating or Go's templating capabilites) vs benefits it brings.

furkatgofurov7 (approved):

PR #434, opened 2024-09-09
Avoid deleting etcd member before node is drained

### LINKED ISSUE #431: Rolling upgrades are blocked by nodes that are not properly drained
**What happened:**

The issue has been described in https://gitlab.com/sylva-projects/sylva-core/-/issues/1595

During rolling upgrades, control-plane node machine is removed from etcd cluster as soon as the machine is being rolled out [here](https://github.com/rancher/cluster-api-provider-rke2/blob/v0.6.0/controlplane/internal/controllers/scale.go#L154-L166).

The issue is that in rke2 deployments, kubelet is configured to use local api server (127.0.0.1:443), which in turn relies on local etcd pod. But as this node is removed from etcd cluster, kubelet won't be able to reach the API any more, and it will fail to properly drain the node as all pods will remain stuck in Terminating state from kubernetes perspective.

We should probably try to avoid removing etcd member so early during rolling upgrades, we could instead rely on periodic [reconcileEtcdMembers](https://github.com/rancher/cluster-api-provider-rke2/blob/v0.6.0/controlplane/internal/controllers/rke2controlplane_controller.go#L511-L515) that ensures the number of etcd members is in sync with the number of machines/nodes, this way etcd members will be removed only after the node has been properly drained and removed from cluster by capi controller.

zioc:
Control plane machines should not be removed from etcd cluster until the node is drained, otherwise kubelet would loose connectivity to the API, as it relies on local api-server, which in turn relies on local etcd pod.

If this etcd member is removed from the cluster, kubelet won't be able to report status to the API, and node won't be properly drained.

The cleanup of removed node will be done by periodic call of [reconcileEtcdMembers](https://github.com/rancher/cluster-api-provider-rke2/blob/967b2da50d7869d08da4b1adf6b090599cfe9c16/controlplane/internal/controllers/rke2controlplane_controller.go#L513) once the node is removed from api by capi controller.

We've observed significant improvements while testing the proposed change in https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/2845.

kind/bug
fixes #431

### DIFF
--- a/controlplane/internal/controllers/scale.go
+++ b/controlplane/internal/controllers/scale.go
@@ -151,20 +151,6 @@ func (r *RKE2ControlPlaneReconciler) scaleDownControlPlane(
 		return ctrl.Result{}, errors.New("failed to pick control plane Machine to delete")
 	}
 
-	// If etcd leadership is on machine that is about to be deleted, move it to the newest member available.
-	etcdLeaderCandidate := controlPlane.Machines.Newest()
-	if err := r.workloadCluster.ForwardEtcdLeadership(ctx, machineToDelete, etcdLeaderCandidate); err != nil {
-		logger.Error(err, "Failed to move leadership to candidate machine", "candidate", etcdLeaderCandidate.Name)
-
-		return ctrl.Result{}, err
-	}
-
-	if err := r.workloadCluster.RemoveEtcdMemberForMachine(ctx, machineToDelete); err != nil {
-		logger.Error(err, "Failed to remove etcd member for machine")
-
-		return ctrl.Result{}, err
-	}
-
 	logger = logger.WithValues("machine", machineToDelete)
 	if err := r.Client.Delete(ctx, machineToDelete); err != nil && !apierrors.IsNotFound(err) {
 		logger.Error(err, "Failed to delete control plane machine")


controlplane/internal/controllers/scale.go
```diff
@@ -151,20 +151,6 @@ func (r *RKE2ControlPlaneReconciler) scaleDownControlPlane(
 		return ctrl.Result{}, errors.New("failed to pick control plane Machine to delete")
 	}
 
-	// If etcd leadership is on machine that is about to be deleted, move it to the newest member available.
```
Danil-Grigorev, line 154: Potentially this change can also check that more than 1 CP machine is not currently tagged for deletion, and perform immediate etcd leadership transfer only on scaling down to 1 CP. Then the regular memebership removal can be done in a [periodic](https://github.com/rancher/cluster-api-provider-rke2/blob/v0.6.0/controlplane/internal/controllers/rke2controlplane_controller.go#L511-L515). But I can’t tell if the leader removed on >2 replicas can be transferred at the moment of scale down to 1.

We should be testing this scenario in e2e, unless there were changes.

zioc: You're right, it would probably be preferable to keep the immediate etcd leadership transfer part, as it'll prevent transcient Leader Failures documented [here](https://github.com/improbable-eng/etcd-cluster-operator/blob/master/docs/operations.md#scale-down-operations).

There could indeed be some regression on https://github.com/rancher/cluster-api-provider-rke2/issues/263, but on the other hand, I tend to believe that this issue is way less critical than having rolling updates of control planes blocked by node drain. 

To that extent, current e2e test are not representative of typical deployments, because as nodeDrainTimeout is [set to 30s](https://github.com/rancher/cluster-api-provider-rke2/blob/main/test/e2e/data/infrastructure/cluster-template-docker.yaml#L119), the described issue has no chance to appear, as nodes will be deleted quickly by capi controller. But in that case we give only few changes to statefull applications to handover in a clean way, and podDisruptionBudgets won't be honored.

That being said, from what I understand, etcd will become read-only when control place is scaled down from 2 to 1 as it will loose its chorum. But in that case, API server should be still available, couldn't we just figure ou a way to ensure that rke2-controller will succes to remove the member corresponding to the deleted machine anyway in that case?

zioc: While searching for pre-terminate hook in cluster-api code, I found that PR that landed a few days ago: https://github.com/kubernetes-sigs/cluster-api/pull/11137.
It seems to be implementing something similar to the second proposition above, it could be a nice example to follow!

zioc: Closing that PR in flavor of https://github.com/rancher/cluster-api-provider-rke2/pull/435 as it is adressing the issue in a better way (following https://github.com/kubernetes-sigs/cluster-api/pull/11137 as discussed above)

PR #435, opened 2024-09-10
Implement etcd member management in pre-terminate hook

### LINKED ISSUE #431: Rolling upgrades are blocked by nodes that are not properly drained

Danil-Grigorev:

Fixes #431 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controlplane/internal/controllers/rke2controlplane_controller.go
+++ b/controlplane/internal/controllers/rke2controlplane_controller.go
@@ -550,21 +550,6 @@ func (r *RKE2ControlPlaneReconciler) reconcileNormal(
 	desiredReplicas := int(*rcp.Spec.Replicas)
 
 	switch {
-	case numMachines == desiredReplicas:
-		nonDeleteingMachines := controlPlane.Machines.Filter(collections.Not(collections.HasDeletionTimestamp))
-		for _, machine := range nonDeleteingMachines {
-			annotaions := machine.GetAnnotations()
-
-			if _, found := annotaions[controlplanev1.PreTerminateHookCleanupAnnotation]; !found {
-				annotaions[controlplanev1.PreTerminateHookCleanupAnnotation] = ""
-				machine.SetAnnotations(annotaions)
-			}
-		}
-
-		// Patch machine annoations
-		if err := controlPlane.PatchMachines(ctx); err != nil {
-			return ctrl.Result{}, err
-		}
 	// We are creating the first replica
 	case numMachines < desiredReplicas && numMachines == 0:
 		// Create new Machine w/ init
@@ -730,7 +715,7 @@ func (r *RKE2ControlPlaneReconciler) reconcileDelete(ctx context.Context,
 
 		// During RKE2CP deletion we don't care about forwarding etcd leadership or removing etcd members.
 		// So we are removing the pre-terminate hook.
-		// This is important because when deleting KCP we will delete all members of etcd and it's not possible
+		// This is important because when deleting RKE2CP we will delete all members of etcd and it's not possible
 		// to forward etcd leadership without any member left after we went through the Machine deletion.
 		// Also in this case the reconcileDelete code of the Machine controller won't execute Node drain
 		// and wait for volume detach.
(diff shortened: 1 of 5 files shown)


Danil-Grigorev:
It seems the previous problem is fixed, but another problem appears on removal of the last machine.
```
I0910 14:37:55.029275       1 machine_controller.go:357] "Skipping deletion of Kubernetes Node associated with Machine as it is not allowed“
```
There is a discussion upstream about a potentially related issue https://kubernetes.slack.com/archives/C8TSNPY4T/p1725952675583209

furkatgofurov7:
> It seems the previous problem is fixed, but another problem appears on removal of the last machine.
> 
> ```
> I0910 14:37:55.029275       1 machine_controller.go:357] "Skipping deletion of Kubernetes Node associated with Machine as it is not allowed“
> ```
> 
> There is a discussion upstream about a potentially related issue https://kubernetes.slack.com/archives/C8TSNPY4T/p1725952675583209

Looks like a potential fix upstream will be available to use with the new v1.8.3 CAPI patch release scheduled for today, however we have not yet bumped CAPI to v1.8.x series in CAPRKE2

Danil-Grigorev:
Watch logs and metric collection causes failures in CI

Danil-Grigorev:
This PR depends on #440

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -698,14 +705,31 @@ func (r *RKE2ControlPlaneReconciler) reconcileDelete(ctx context.Context,
...
 
 	var errs []error
 
 	for i := range machinesToDelete {
 		m := machinesToDelete[i]
 		logger := logger.WithValues("machine", m)
 
+		// During RKE2CP deletion we don't care about forwarding etcd leadership or removing etcd members.
+		// So we are removing the pre-terminate hook.
+		// This is important because when deleting KCP we will delete all members of etcd and it's not possible
```
tmmorin, line 718:
```suggestion
		// This is important because when deleting RKE2CP we will delete all members of etcd and it's not possible
```

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -909,6 +935,118 @@ func (r *RKE2ControlPlaneReconciler) ClusterToRKE2ControlPlane(ctx context.Conte
...
+		if etcdLeaderCandidate != nil {
+			if err := workloadCluster.ForwardEtcdLeadership(ctx, deletingMachine, etcdLeaderCandidate); err != nil {
+				return ctrl.Result{}, errors.Wrapf(err, "failed to move leadership to candidate Machine %s", etcdLeaderCandidate.Name)
+			}
+		} else {
+			log.Info("Skip forwarding etcd leadership, because there is no other control plane Machine without a deletionTimestamp")
+		}
+
+		// Note: Removing the etcd member will lead to the etcd and the kube-apiserver Pod on the Machine shutting down.
+		// If ControlPlaneKubeletLocalMode is used, the kubelet is communicating with the local apiserver and thus now
```
tmmorin, line 1023: isn't this comment kubeadm-specific ?

Danil-Grigorev: Yes, this comment can be simplified.

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -909,6 +935,116 @@ func (r *RKE2ControlPlaneReconciler) ClusterToRKE2ControlPlane(ctx context.Conte
...
+
+	// Return early because the Machine controller is not yet waiting for the pre-terminate hook.
+	c := conditions.Get(deletingMachine, clusterv1.PreTerminateDeleteHookSucceededCondition)
+	if c == nil || c.Status != corev1.ConditionFalse || c.Reason != clusterv1.WaitingExternalHookReason {
+		return ctrl.Result{RequeueAfter: deleteRequeueAfter}, nil
+	}
+
+	// The following will execute and remove the pre-terminate hook from the Machine.
+
+	// If we have more than 1 Machine and etcd is managed we forward etcd leadership and remove the member
```
alexander-demicev, line 1001: do we support not-managed etcd?

Danil-Grigorev: No, the code reflects that.

alexander-demicev (approved):
thanks a lot for taking care of this issue

furkatgofurov7 (approved):
Thanks for the fix @Danil-Grigorev!

PR #436, opened 2024-09-11
bootstrap: switch to chrony for ignition config

### LINKED ISSUE #430: NTP API broken for Leap/SLEMicro
**What happened:**

We provide an [API for NTP configuration](https://github.com/rancher/cluster-api-provider-rke2/blob/main/bootstrap/api/v1beta1/rke2config_types.go#L88) which influences the firstboot configuration to configure/enable  a NTP client.

In the `format: ignition` case we can see the [relevant templating](https://github.com/rancher/cluster-api-provider-rke2/blob/main/bootstrap/internal/ignition/butane/butane.go#L65) e.g:

```
./butane/butane.go:    {{- if .NTPServers }}
./butane/butane.go:    - name: ntpd.service
./butane/butane.go:    {{- if .NTPServers }}
```

However our primary target for ignition support is Leap/SLEMicro, which doesn't include ntpd, it uses chrony instead, so we see an ignition error like:

`Failed to enable unit, unit ntpd.service does not exist.`

We should configure/enable chrony instead as described in [the SLEMicro docs](https://documentation.suse.com/sle-micro/6.0/html/Micro-ntp-time-synchronization/index.html#ntp-configure-configuration-file)

**How to reproduce it:**
Deploy with NTP servers defined/enabled - observe error and that the OS client is not configured

**Environment:**
- rke provider version: 0.6.1
- OS : Leap/SLEMicro 5.5/6.0

hardys:
Leap/SLEMicro don't include ntpd so the current approach doesn't work

kind/bug

Currently the `ntp.servers` API provided e.g by `RKE2ControlPlane` does not work for Leap/SLEMicro, because chrony is used not ntpd as currently configured.

Fixes #430

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/bootstrap/internal/ignition/butane/butane.go
+++ b/bootstrap/internal/ignition/butane/butane.go
@@ -34,12 +34,12 @@ import (
 )
 
 // The template contains configurations for two main sections: systemd units and storage files.
-// The first section defines two systemd units: rke2-install.service and ntpd.service.
+// The first section defines two systemd units: rke2-install.service and chronyd.service.
 // The rke2-install.service unit is enabled and is executed only once during the boot process to run the /etc/rke2-install.sh script.
 // This script installs and deploys RKE2, and performs pre and post-installation commands.
-// The ntpd.service unit is enabled only if NTP servers are specified.
+// The chronyd.service unit is enabled only if NTP servers are specified.
 // The second section defines storage files for the system. It creates a file at /etc/rke2-install.sh.
-// If NTP servers are specified, it creates an NTP configuration file at /etc/ntp.conf.
+// If NTP servers are specified, it creates an NTP configuration file at /etc/chrony.conf.
 const (
 	butaneTemplate = `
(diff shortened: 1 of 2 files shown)


hardys:
While making these changes I also noticed that the [ntp.enabled flag](https://github.com/rancher/cluster-api-provider-rke2/blob/main/bootstrap/api/v1beta1/rke2config_types.go#L195) doesn't seem to work (for the cloud-init and ignition cases) - I didn't try to fix that here as the flag seems largely redundant, but it may warrant a follow up issue?

alexander-demicev (approved):
thanks!

furkatgofurov7 (approved):

Danil-Grigorev (approved):
Hi @hardys, thanks for the fix! Would you be fine to rebase on `main`? We suspect the failures in CI are not related to this PR and it should fix it.

PR #450, opened 2024-10-01
🐛 Prevent machine re-rollout on system agent start command

Danil-Grigorev:

This agent installation command is not executed during bootstrap process, due to `sudo` prefix. Changing it to a correct command, causes machines to re-rollout after successful bootstrap endlessly.

Part of https://github.com/rancher/turtles/issues/763

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/rke2/machine_filters.go
+++ b/pkg/rke2/machine_filters.go
@@ -76,7 +76,7 @@ func matchesRKE2BootstrapConfig(machineConfigs map[string]*bootstrapv1.RKE2Confi
 			cmds := []string{}
 
 			for _, cmd := range machineConfig.Spec.PostRKE2Commands { // Filter out commands that are injected by the Rancher Turtles webhook
-				if cmd == "sudo sh /opt/system-agent-install.sh" {
+				if cmd == "sh /opt/system-agent-install.sh" {
 					continue
 				}
 


pkg/rke2/machine_filters.go
```diff
@@ -76,7 +76,7 @@ func matchesRKE2BootstrapConfig(machineConfigs map[string]*bootstrapv1.RKE2Confi
 			cmds := []string{}
 
 			for _, cmd := range machineConfig.Spec.PostRKE2Commands { // Filter out commands that are injected by the Rancher Turtles webhook
-				if cmd == "sudo sh /opt/system-agent-install.sh" {
+				if cmd == "sh /opt/system-agent-install.sh" {
```
alexander-demicev, line 79: before merging this, can we figure out why system-agent install script requires root permissions to run https://github.com/rancher/system-agent/blob/main/install.sh#L886?

furkatgofurov7 (approved):

pkg/rke2/machine_filters.go
Danil-Grigorev: To add and enable systemd unit with a binary running as root you would need high privileges. Bootstrap commands already at root, so this works.

alexander-demicev (approved):

Issue #449, opened 2024-09-30
ETCD becomes unavailable on update


alexander-demicev:
See https://gitlab.com/sylva-projects/sylva-core/-/issues/1687 for all details describing the issue

tmmorin:
thanks for opening this issue

one precision: the symptom is less clear that "etcd isn't available"... what we clearly see is some controllers failing to get leases from the apiServer, but some controller apparently still can, and we aren't sure that reading from the api is an issue

there are "bad certificates" issues in etcd logs, but they possibly are harmless (they relate to a node which isn't anymore an etcd cluster member)

so at this point we aren't really sure what the best issue title is ... ;-)

tmmorin:
here are two dumps of CI jobs that exhibited the issue:
* https://gitlab.com/sylva-projects/sylva-core/-/jobs/7951588127/artifacts/download (on capo)
* https://gitlab.com/-/project/42451983/uploads/b173fc74cea9cf2d9596e8bb55bd3aff/artifact-7915786264.zip (on capm3)

The relevant info is in management-cluster-dump directory, it has the usual per-ns kubectl cluster-info dumps with pod logs, but also resource dumps for many CRDs, it also has a "clusterctl-describe.txt" which is convenient to see what was CAPI state.

tmmorin:
To have ideas on whether or not our suspicion that https://github.com/rancher/cluster-api-provider-rke2/pull/404 could be contributing to this issue, since yesterday I've ran our CI pipelines with this commit reverted.   In the dozen runs that I did with this commit reverted, I could not reproduce this issue a single time.

This is no proof though, of course, but may help root cause the problem.

We suspected this commit to be a possible cause base on this plausible story line:
- (errors in our pipelines related to etcd certificates - `bad certificate`)
- the fact that we upgrade from a starting point not having the cluster-etcd secret
- ... to a setup where the RKE2 bootstrap provider adds the working kubectl that creates the cluster-etcd secret
- the fact that issue arise right after the creation of the secret (rebuild of the first node with the newer RKE2 bootstrap provider)
We're thinking that we're possibly in a state where the upgrade path designed in https://github.com/rancher/cluster-api-provider-rke2/pull/265 would have some defect.

Danil-Grigorev:
[...]

PR #453, opened 2024-10-04
Allow to distinguish legacy CP without ETCD management via annotation

Danil-Grigorev:

This change allows to disable ETCD certificate management and provisioning with CAPI generated certificates for legacy control planes via `controlplane.cluster.x-k8s.io/legacy` annotation.

This annotation is meant to be set on RKE2 CP, upgraded from 0.2.z to version 0.3+. When the child cluster does not have `cluster-etcd` secret created, and the `<cluster-name>-etcd` secret in the management cluster does not contain the external `cluster.x-k8s.io/purpose` label, then this cluster requires the annotation to be set.

This would ensure that automatically generated ETCD certificates by CAPI will not be used for new ETCD replicas in provisioned nodes, replacing existing set and creating 2 separate etcd groups.

This change is a followup to #449 

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/bootstrap/internal/controllers/rke2config_controller.go
+++ b/bootstrap/internal/controllers/rke2config_controller.go
@@ -369,6 +369,10 @@ func (r *RKE2ConfigReconciler) handleClusterNotInitialized(ctx context.Context, 
 	}()
 
 	certificates := secret.NewCertificatesForInitialControlPlane()
+	if _, found := scope.ControlPlane.Annotations[controlplanev1.LegacyRKE2ControlPlane]; found {
+		certificates = secret.NewCertificatesForLegacyControlPlane()
+	}
+
 	if err := certificates.LookupOrGenerate(
 		ctx,
 		r.Client,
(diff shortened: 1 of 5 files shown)


Danil-Grigorev:
To perform certificate rotation https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates guide can be used.

In RKE2 case that would translate into requirement to collect certificates from the node filesystem, located under `/var/lib/rancher/rke2/server/tls/etcd` into appropriate secrets on the management cluster.

In case of ETCD, the `server-ca.crt` and `server-ca.key` should be stored in the secret under `tls.crt` and `tls.key` on the management cluster inside the `<cluster-name>-etcd` secret. This should convert cluster from `legacy` to regular, after which the label can be removed from the RKE2 CP object.

Danil-Grigorev:
After some testing, it seems that upgrade from 0.2.7 to latest is safe without this change. A set of certificates is only passed on bootstrap to the nodes in case of `ControlPlaneInitialized` condition not being true on the cluster, so the pre-GA clusters should be safe, even with a different set of certificates stored in `<cluster-name>-<purpose>` secrets.

furkatgofurov7 (approved):

alexander-demicev (approved):

PR #481, opened 2024-10-24
Use REGISTRY env var based on the actual image pushing registry expected

furkatgofurov7:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Build docker image for ghcr.io
       run: make docker-build-all TAG=${{ env.TAG }}
     - name: Push docker image to ghcr.io
-      run: make docker-push-all TAG=${{ env.TAG }} PROD_REGISTRY=${{ env.GHCR_REGISTRY }}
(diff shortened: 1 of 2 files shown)


Danil-Grigorev (approved):

.github/workflows/release.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         go-version: '=1.21.8'
     - name: Update manifests
       run: |
-        make release RELEASE_TAG=${{ env.TAG }}
+        make release RELEASE_TAG=${{ env.TAG }} REGISTRY=${{ env.GHCR_REGISTRY }}
```
furkatgofurov7, line 72: NOTE: I am setting this to ghcr and not the prime registry to keep the previous behaviour. If not set explicitly here like we had before, it will use `PRIME-REGISTRY-REGISTRY` for `REGISTRY` since it is the recent value  set to the `REGISTRY` env var

alexander-demicev (approved):

PR #483, opened 2024-10-25
feat: Improvements to make image buildings faster

furkatgofurov7:
Support only amd64/armd64 archs and publish images to speed up the image building process. We might consider adding other archs in the future based on the need and demand from the users.

To further speed up build process, this patch removes redundant steps from Dockerfile. The mod-related steps are redundant because go build handles downloading dependencies as needed already and the existing build step is already configured to cache them, so there shouldn't be any meaningful difference in build times when dependencies are added/updated.

Last, we bump go version to v1.22.x to keep it more or less newer.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,7 @@
     // Features to add to the dev container. More info: https://containers.dev/features.
     "features": {
         "ghcr.io/devcontainers/features/go:1": {
-            "version": "1.21"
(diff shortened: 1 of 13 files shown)


alexander-demicev (approved):

salasberryfin (approved):
Thanks @furkatgofurov7

yiannistri (approved):

PR #542, opened 2025-01-06
Deprecate infrastructure ref

### LINKED ISSUE #341: Can not create Cluster from Cluster Class without repeating KCP's infrastructureRef
**What happened:**
When defining my RKE2 cluster class I configured the `RKE2ControlPlaneTemplate` as follows:

```yaml
apiVersion: controlplane.cluster.x-k8s.io/v1beta1
kind: RKE2ControlPlaneTemplate
metadata:
  name: rke2-control-plane
  namespace: default
spec:
  template:
    spec:
      nodeDrainTimeout: 2m
      registrationMethod: "control-plane-endpoint"
      rolloutStrategy:
        rollingUpdate:
          maxSurge: 1
        type: RollingUpdate
      serverConfig:
        disableComponents:
          kubernetesComponents:
            - cloudController
```

In the CC definition the `spec.controlPlane.machineInfrastructure.ref` is correctly set:

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: ClusterClass
metadata:
  name: rke2
  namespace: default
spec:
  controlPlane:
    machineInfrastructure:
      ref:
        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
        kind: ElementalMachineTemplate
        name: rke2-control-plane
    ref:
      apiVersion: controlplane.cluster.x-k8s.io/v1beta1
      kind: RKE2ControlPlaneTemplate
      name: rke2-control-plane
```

However RKE2 provider fails to initialize the control plane machines:

```
I0531 08:47:34.318733       1 rke2controlplane_controller.go:430] "Reconcile RKE2 Control Plane" controller="rke2controlplane" controllerGroup="controlplane.cluster.x-k8s.io" controllerKind="RKE2ControlPlane" RKE2ControlPlane="default/rke2-clusterclass-dfkb5" namespace="default" name="rke2-clusterclass-dfkb5" reconcileID="ceadb0a6-1099-43d6-a09e-ea2f21304cb9"
I0531 08:47:34.329520       1 rke2controlplane_controller.go:549] "Initializing control plane" controller="rke2controlplane" controllerGroup="controlplane.cluster.x-k8s.io" controllerKind="RKE2ControlPlane" RKE2ControlPlane="default/rke2-clusterclass-dfkb5" namespace="default" name="rke2-clusterclass-dfkb5" reconcileID="ceadb0a6-1099-43d6-a09e-ea2f21304cb9" Desired=1 Existing=0
E0531 08:47:34.329689       1 scale.go:74] "Failed to create initial control plane Machine" err="failed to clone infrastructure template: failed to retrieve  external object \"default\"/\"\": Object 'Kind' is missing in 'unstructured object has no kind'" namespace="default" name="rke2-clusterclass-dfkb5" cluster-name="rke2-clusterclass"
```

The workaround is to repeat the `infrastructureRef` in the `RKE2ControlPlaneTemplate`, like this:

```yaml
apiVersion: controlplane.cluster.x-k8s.io/v1beta1
kind: RKE2ControlPlaneTemplate
metadata:
  name: rke2-control-plane
  namespace: default
spec:
  template:
    spec:
      infrastructureRef:
        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
        kind: ElementalMachineTemplate
        name: rke2-control-plane
      nodeDrainTimeout: 2m
      registrationMethod: "control-plane-endpoint"
      rolloutStrategy:
        rollingUpdate:
          maxSurge: 1
        type: RollingUpdate
      serverConfig:
        disableComponents:
          kubernetesComponents:
            - cloudController
```

I think this should not be necessary due to the CC definition.

The RKE2ControlPlaneTemplate should respect what was defined in the Cluster Class.

**How to reproduce it:**

See config above, or it can be reproduced with the quickstart sample by removing the `RKE2ControlPlaneTemplate.spec.template.spec.infrastructureRef` object. [reference](https://github.com/rancher-sandbox/cluster-api-provider-rke2/blob/main/samples/docker/clusterclass/clusterclass-quick-start.yaml#L134) 

**Environment:**

- rke provider version: main branch

anmazzotti:
kind/feature
kind/bug

This is a (dirty) attempt of using the v1beta1 newly added `RKE2ControlPlane.Spec.MachineTemplate.InfrastructureRef` and fixes the API to be used properly with ClusterClass.

Note that this counts as an API change, however this is an attempt of getting away with it without a version bump.
The controlplane controller tries to catch and patch all missed conversions for this reason.

Still a draft, missing proper testing and test coverage.

Fixes #341

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


controlplane/api/v1alpha1/conversion.go
```diff
@@ -184,9 +184,57 @@ func (dst *RKE2ControlPlaneTemplateList) ConvertFrom(srcRaw conversion.Hub) erro
 }
 
 func Convert_v1beta1_RKE2ControlPlaneSpec_To_v1alpha1_RKE2ControlPlaneSpec(in *controlplanev1.RKE2ControlPlaneSpec, out *RKE2ControlPlaneSpec, s apiconversion.Scope) error {
-	// Version was added in v1beta1.
-	// MachineTemplate was added in v1beta1.
-	return autoConvert_v1beta1_RKE2ControlPlaneSpec_To_v1alpha1_RKE2ControlPlaneSpec(in, out, s)
```
Danil-Grigorev, line 189: We shouldn’t change the generated v1alpha1 conversion code for a change introduced in v1beta1 API, compatible with previous version. It additionally does not cover v1beta1 API state, when the ref is not set correctly and there is no v1alpha1 resource version. I think these changes could be better placed in a defaulting webhook.

anmazzotti: The conversion needs to be implemented correctly, as they meant to be, for all users still on the previous version.

The current v1beta1 "conversion" is done by the rke2 control plane controller within the same PR. It's the quirky part. Not sure the webhook will work, we must ensure that any existing RKE2ControlPlane resource is "sanitized"/"converted" before it's reconciled.

controlplane/api/v1beta1/rke2controlplane_types.go
```diff
@@ -72,11 +71,14 @@ type RKE2ControlPlaneSpec struct {
 
 	// InfrastructureRef is a required reference to a custom resource
 	// offered by an infrastructure provider.
+	// This field is deprecated. Use `.machineTemplate.infrastructureRef` instead.
```
alexander-demicev, line 74: we can add `// +kubebuilder:deprecatedversion` tag
https://book.kubebuilder.io/reference/markers/crd

controlplane/api/v1alpha1/conversion.go
Danil-Grigorev: Yes, but in this case we do not have API changes which require altering conversion code, unless there is an issue (if any) with the correctness of the current implementation. In which case it is better addressed as a separate issue/PR.

The defaulting webhook could be a good place to modify the resource state before it’s reconciled, while in the code we can just return early to invoke the `patch` on the RKE2CP or double make sure that the field is populated (as it is now).

salasberryfin:
@anmazzotti looks like this needs a rebase

controlplane/api/v1beta1/rke2controlplane_types.go
```diff
@@ -58,7 +58,6 @@ type RKE2ControlPlaneSpec struct {
 
 	// MachineTemplate contains information about how machines
 	// should be shaped when creating or updating a control plane.
-	// +optional
```
Danil-Grigorev, line 61: I believe the `machineTemplate` is [optional](https://github.com/kubernetes-sigs/cluster-api/blob/7f2a8cdd5524692e6a4554c01716bba61cd10490/controlplane/kubeadm/api/v1beta1/kubeadmcontrolplanetemplate_types.go#L75) in kubeadm.

anmazzotti: Feels like an issue in kubeadm to me, also being a pointer there.

I can change it for the sake of alignment, but nothing will work without machineTemplate, it should not be optional.

anmazzotti: And for reference this is how it was implemented in k3s: https://github.com/k3s-io/cluster-api-k3s/blob/main/controlplane/api/v1beta2/kthreescontrolplane_types.go#L82

However I'm also the author of that, so it's biased. :D

Danil-Grigorev: I opened an issue: https://github.com/rancher/cluster-api-provider-rke2/issues/554 to track API changes, including marking this field as `required`. Feel free to add or change anything, to shape it better.

I’d prefer not changing it here, since making field `required` is a breaking API change.

Danil-Grigorev (approved):
LGTM, thanks!

furkatgofurov7:
@anmazzotti I am afraid this PR needs a rebase

controlplane/api/v1beta1/rke2controlplane_webhook.go
```diff
@@ -48,6 +48,16 @@ var _ webhook.Defaulter = &RKE2ControlPlane{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (r *RKE2ControlPlane) Default() {
 	bootstrapv1.DefaultRKE2ConfigSpec(&r.Spec.RKE2ConfigSpec)
+
+	// (Quirk)
```
furkatgofurov7, line 52: can we remove the comments?

furkatgofurov7 (dismissed):
LGTM

controlplane/api/v1beta1/rke2controlplane_webhook.go
Danil-Grigorev:
```suggestion
	// Defaults MachineTemplate.InfrastructureRef to Spec.InfrastructureRef, if missing
```
seems to be more formal

controlplane/api/v1beta1/rke2controlplane_types.go
Danil-Grigorev: One more nit: since we currently need to have either `Spec.MachineTemplate.InfrastructureRef` or `Spec.InfrastructureRef` populated at any moment without breaking changes, it can also be covered in the validating webhook for completeness. Something like:

```go
	if r.Spec.MachineTemplate.InfrastructureRef.Name == “” && r.Spec.InfrastructureRef.Name == "" {
		allErrs = append(allErrs,
			field.Invalid(field.NewPath("spec", "machineTemplate", "infrastructureRef"),
				r.Spec.MachineTemplate.InfrastructureRef, "machineTemplate is required"))
	}
```

salasberryfin (dismissed):

Danil-Grigorev (approved):

furkatgofurov7 (approved):

Danil-Grigorev (approved):

PR #583, opened 2025-02-25
Fix cloud provider installation in AWS template

alexander-demicev:

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/examples/aws/cluster-template.yaml
+++ b/examples/aws/cluster-template.yaml
@@ -123,6 +123,9 @@ spec:
   - sudo hostnamectl set-hostname $(curl -s http://169.254.169.254/1.0/meta-data/hostname)
   agentConfig:
     airGapped: true
+    kubelet:
+      extraArgs:
+      - --cloud-provider=external
   machineTemplate:
     infrastructureRef:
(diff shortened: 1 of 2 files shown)


anmazzotti (dismissed):

furkatgofurov7 (dismissed):

examples/aws/cluster-template.yaml
```diff
@@ -229,6 +235,7 @@ apiVersion: addons.cluster.x-k8s.io/v1beta1
 kind: ClusterResourceSet
 metadata:
   name: crs-ccm
+  namespace: ${NAMESPACE}
```
Danil-Grigorev, line 238: I think when specifying `-n <namespace>` in `kubectl apply`, namespace is being set for each resource from the list. It may however conflict with an already set value. Could this value be removed from templates instead?

alexander-demicev: @Danil-Grigorev the namespace if there to align with `clusterctl` templates, where it's used to substitute namespace in `generate` command

furkatgofurov7 (approved):

anmazzotti (approved):

PR #603, opened 2025-03-20
added openstack cluster example

djcenox:
kind/documentation
kind/regression
kind/*

Just an example of an openstack cluster with rke2.
If using openstack LB check out the label on controlplane: **node.kubernetes.io/exclude-from-external-load-balancers** that is missing, it needs to be added manually.

Fixes #
Nothing 

**Checklist**:

### DIFF
--- a/docs/book/src/01_user/01_getting-started.md
+++ b/docs/book/src/01_user/01_getting-started.md
@@ -53,7 +53,7 @@ In order to use this provider, you need to have a management cluster available t
 No additional steps are required and you can install the RKE2 provider with **clusterctl** directly:
 
 ```bash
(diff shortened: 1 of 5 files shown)


examples/templates/openstack/cluster-template.yaml
```diff
@@ -0,0 +1,174 @@
...
+  template:
+    metadata:
+      labels:
+        cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
+    spec:
+      bootstrap:
+        configRef:
+          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+          kind: RKE2ConfigTemplate
+          name: ${CLUSTER_NAME}-workers
```
anmazzotti, line 61: Should the `RKE2ConfigTemplate` also be included in this manifest?

djcenox: Yes definitely, addind asap, sorry.

PR #647, opened 2025-05-12
Ensure RKE2Controlplane machine labels match template labels

### LINKED ISSUE #616: Remove the `SkipMachineLabelsCheck` boolean from e2e tests
Undo the changes done in https://github.com/rancher/cluster-api-provider-rke2/pull/611/. Specifically in the commit https://github.com/rancher/cluster-api-provider-rke2/pull/611/commits/e664e98892c60d2d1f620d9f4bed374f5b374398.

These changes were made because v0.13.0 would not have the latest changes being made in that PR. Hence, the upgrade test was failing.

cc @anmazzotti

furkatgofurov7:

Fixes #616 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -63,7 +63,6 @@ type ApplyClusterTemplateAndWaitInput struct {
 	Args                         []string
 	PreWaitForCluster            func()
 	PostMachinesProvisioned      func()
-	SkipMachineLabelCheck        bool
 	ControlPlaneWaiters
 }
 
@@ -80,7 +79,6 @@ type ApplyCustomClusterTemplateAndWaitInput struct {
 	Args                         []string
 	PreWaitForCluster            func()
 	PostMachinesProvisioned      func()
-	SkipMachineLabelCheck        bool
 	ControlPlaneWaiters
 }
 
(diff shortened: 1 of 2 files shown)


test/e2e/helpers.go
```diff
@@ -400,21 +394,6 @@ func WaitForRKE2ControlPlaneMachinesToExist(ctx context.Context, input WaitForRK
 		}
 		return count, nil
 	}, intervals...).Should(Equal(int(*input.ControlPlane.Spec.Replicas)), "Timed out waiting for %d control plane machines to exist", int(*input.ControlPlane.Spec.Replicas))
-
-	if !input.SkipMachineLabelCheck {
-		// check if machines owned by RKE2ControlPlane have the labels provided in .spec.machineTemplate.metadata.labels
-		controlPlaneMachineTemplateLabels := input.ControlPlane.Spec.MachineTemplate.ObjectMeta.Labels
```
anmazzotti, line 406: This should not be removed.

anmazzotti (changes requested):
The test was removed but it should stay. 
Anyway, I had to address this in another PR: https://github.com/rancher/cluster-api-provider-rke2/pull/646

In that PR I also need a similar flag to skip some tests, so I generalized it.

test/e2e/helpers.go
anmazzotti: The bool was introduced to skip testing this feature when we do the e2e upgrade test. This test will first install the latest version of RKE2 provider. 
Since the feature was not present in latest version, it was impossible for the test to fail. 
Unfortunately the label was set through false through the entire test (so also after upgrading from latest to main), which was incorrect. 

Anyway, may you please drop this PR in favor of #646 , so we don't create an unnecessary conflict. 

I also replaced the bool with a more generic one, so that it can be exploited in the future for similar cases.

furkatgofurov7: thanks for background information, closing this in favor of #646

PR #684, opened 2025-06-26
Rely on RKE2 managed etcd for member removal

### LINKED ISSUE #655: Machines stuck in Deleting during k8s upgrade
**What happened:**
A k8s upgrade was triggered on a HA (3CP nodes) downstream cluster where we changed the following things in the RK2ControlPlane resource:

- The version was changed from `v1.31.7+rke2r1` to `v1.32.3+rke2r1`
- `rolloutStrategy.rollingUpdate.maxSurge` was changed to `0`

What happened then ->
1. CP node C1 was torn down for a node rolling update
2. Draining was started on C1
3. The RKE2 provider controller has just removed its etcd membership
4. The server, the OS, still hasn't been deleted (CAPI would soon finalize the deletion of the Machine and the infra provider would delete the InfraMachine, but this hasn't happened yet)
5. Since the local etcd on node C1 was down, the Apiserver on node C1 (that only talks to the etcd server on node C1) cannot work 
      - Apiserver pods became NotReady, but remained alive and running
6. The Machine/InfraMachine never gets deleted, and many controllers that need to talk to the API can't talk to the API server context canceled, timeouts.

We expect the RKE2 provider to behave consistently and reliably during node roll-out operations, maintaining proper coordination with the nodes throughout the process.

**How to reproduce it:**
Described above.
**Note**: This issue is somewhat intermittent and does not always occur because:
1. As long as the CAPI controller and the infrastructure provider controller maintain a connection to node C2 and node C3, they continue functioning normally.
2. In such cases, the server or OS for node C1 is typically shut down quickly, which instantly resolves the issue.
3. Additionally, it appears that the rke2 process occasionally restarts, possibly restarting the Apiserver pod. While this behavior is not fully confirmed, it may reset stale TCP connections, further contributing to the resolution.

**Environment:**
RKE2 MNGM cluster with Metal3 and CAPI and HA Downstream cluster

- rke provider version: v0.15.1
- OS (e.g. from `/etc/os-release`): slm6.1

anmazzotti:
A few changes regarding etcd reconciliation with this PR.

The summary is that caprke2 no longer tries to reconcile etcd member removal. 
The existing logic of removing the member works with etcd, but makes the rke2-server panic when there are no more voters on a node, creating the scenario for a known issue: https://github.com/rancher/rke2/issues/2829

Comparing with the Rancher provisioning API, seems that RKE2 has a different "managed" mechanism for etcd member removal, and that is to simply apply the `etcd.rke2.cattle.io/remove` annotation on the Node. 
After the member has been successfully removed, the `etcd.rke2.cattle.io/removed-node-name` annotation will be populated.
Reference: https://github.com/rancher/rancher/blob/release/v2.11/pkg/controllers/capr/etcdmgmt/saferemoval.go

caprke2 is now implementing the same logic.

Additionally, I removed all reconciliation code for etcd.
This logic ensured that if a Node was deleted, the related etcd member was also being removed.

However this is redundant with what RKE2 already does, so I simply removed the code.
Reference: https://github.com/k3s-io/k3s/blob/release-1.33/pkg/etcd/member_controller.go#L109

I left the "forward etcd leadership" logic in place. It's technically not necessary, but it makes rollouts smoother, since the leader is elected arbitrarily using the newest control plane node.

The pivot test is still flaky. 
On GitHub it fails consistently, during `clusterctl init` the workload cluster. Maybe using a [better runner](https://github.com/rancherlabs/eio/issues/3460) will improve this behavior.
Additionally, CAPD seems to be hanging from time to time. This also has to be [solved](https://github.com/rancher/cluster-api-provider-rke2/issues/691) for the test to be enabled. 
The test can be ran with `GINKGO_FOCUS="Pivot" GINKGO_SKIP="" make test-e2e`
If it gets stuck, the container simply needs a little kick: `docker start <containerid>`
 

Fixes #655

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -8,26 +8,8 @@ on:
 
 jobs:
   e2e-tests:
-    runs-on: ubuntu-latest
+    runs-on: runs-on,runner=4cpu-linux-x64,run-id=${{ github.run_id }}
     steps:
-      - name: Cleanup space
-        run: |
-          # Cherry-pick from jlumbroso/free-disk-space@main
-          sudo apt-get remove -y '^aspnetcore-.*' || echo "::warning::The command [sudo apt-get remove -y '^aspnetcore-.*'] failed to complete successfully. Proceeding..."
-          sudo apt-get remove -y '^dotnet-.*' --fix-missing || echo "::warning::The command [sudo apt-get remove -y '^dotnet-.*' --fix-missing] failed to complete successfully. Proceeding..."
-          sudo apt-get remove -y '^llvm-.*' --fix-missing || echo "::warning::The command [sudo apt-get remove -y '^llvm-.*' --fix-missing] failed to complete successfully. Proceeding..."
-          sudo apt-get remove -y 'php.*' --fix-missing || echo "::warning::The command [sudo apt-get remove -y 'php.*' --fix-missing] failed to complete successfully. Proceeding..."
-          sudo apt-get remove -y '^mongodb-.*' --fix-missing || echo "::warning::The command [sudo apt-get remove -y '^mongodb-.*' --fix-missing] failed to complete successfully. Proceeding..."
-          sudo apt-get remove -y '^mysql-.*' --fix-missing || echo "::warning::The command [sudo apt-get remove -y '^mysql-.*' --fix-missing] failed to complete successfully. Proceeding..."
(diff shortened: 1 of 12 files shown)


.github/workflows/e2e.yml
```diff
@@ -8,26 +8,8 @@ on:
 
 jobs:
   e2e-tests:
-    runs-on: ubuntu-latest
+    runs-on: runs-on,runner=4cpu-linux-x64,run-id=${{ github.run_id }}
```
anmazzotti, line 11: The new self-hosted runner is needed for the increased disk space, since we now `docker save` a few more images for the Pivot test.

salasberryfin (approved):

PR #686, opened 2025-06-27
Revert token registry fix

anmazzotti:
Turns out that #673 was incorrect and needs to be reverted.
The original `identity_token` yaml key was correct.

I followed the wharfie libraries implementation that the RKE2 agent is using to match the AuthConfig struct now.
This is also why the `json` key is `identitytoken` instead of `identity_token` like the `yaml`. See: https://github.com/rancher/wharfie/blob/main/pkg/registries/types.go#L28

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/book/src/02_topics/04_embedded-registry.md
+++ b/docs/book/src/02_topics/04_embedded-registry.md
@@ -1,12 +1,10 @@
 # Configuring Embedded Registry in RKE2
 
 ## Overview
-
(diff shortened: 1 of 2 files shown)


alexander-demicev (approved):

docs/book/src/02_topics/04_embedded-registry.md
```diff
@@ -63,3 +63,4 @@ When configuring the `authSecret`, ensure the secret contains the following keys
 
 - **`username` and `password`** - When using Basic Auth credentials
 - **`identity-token`** - When using a personal access token
```
thehejik, line 65:
```suggestion
- **`identity_token`** - When using a personal access token
```

salasberryfin (approved):

PR #694, opened 2025-07-04
Configures usage of feature-gate & manager flags

### LINKED ISSUE #588: Document and allow usage of managers flags
Almost none of the flags are configurable by the end user, since they miss matching environment variables in the *-components.yaml

Also none of the flag is documented anywhere. 
Applies to both bootstrap and controlplane managers.

apoorvajagtap:
kind/feature
kind/documentation

Allows configuring following feature-gate & manager flags:
- ClusterTopology : featureGate
- namespace : flag
- concurrency  : flag

Fixes #588 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/bootstrap/config/manager/manager.yaml
+++ b/bootstrap/config/manager/manager.yaml
@@ -21,11 +21,12 @@ spec:
       - command:
         - /manager
         args:
-        - --leader-elect
+        - "--leader-elect"
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
-        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
+        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true},ClusterTopology=${CLUSTER_TOPOLOGY:=true}"
+        - "--concurrency=${CONCURRENCY_NUMBER:=10}"
         image: controller:latest
         name: manager
         ports:
--- a/controlplane/config/manager/manager.yaml
+++ b/controlplane/config/manager/manager.yaml
@@ -25,6 +25,7 @@ spec:
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
+        - "--concurrency=${CONCURRENCY_NUMBER:=10}"
         image: controller:latest
         name: manager
         env:
--- a/docs/book/src/02_topics/05_configure-manager-options.md
+++ b/docs/book/src/02_topics/05_configure-manager-options.md
@@ -1,0 +1,46 @@
+## Supported Manager Options
+
+At present, the following feature gates and flags are supported for configuration.
+
+### Feature Gates
+
+| Name            | Description                                           | Default | Env Variable     |
+|-----------------|-------------------------------------------------------|---------|------------------|
+| MachinePool     | for MachinePool functionality                         | true    | EXP_MACHINE_POOL |
+| ClusterTopology | for ClusterClass and managed topologies functionality | true    | CLUSTER_TOPOLOGY |
+
+### Flags
+
+| Name        | Description                                                        | Default | Env Variable       |
+|-------------|--------------------------------------------------------------------|---------|--------------------|
+| concurrency | Number of core resources to process simultaneously                 | 10      | CONCURRENCY_NUMBER |
+
+## Configuring Manager Options
+In order to configure the manager options, it is required to patch the respective values in the 
+`rke2-bootstrap-controller-manager` and `rke2-control-plane-controller-manager` deployments.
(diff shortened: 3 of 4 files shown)


bootstrap/config/manager/manager.yaml
```diff
@@ -21,11 +21,18 @@ spec:
...
+        - "--leader-elect"
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
-        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
+        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true},ClusterTopology=${CLUSTER_TOPOLOGY:=true}"
+        - "--namespace=$(WATCH_NAMESPACE)"
+        - "--concurrency=$(CONCURRENCY_NUMBER)"
+        env:
+        - name: WATCH_NAMESPACE
```
anmazzotti, line 32: These are not needed here. Clusterctl does the substitution on the manifest itself before applying it.

bootstrap/config/manager/manager.yaml
```diff
@@ -21,11 +21,18 @@ spec:
...
         - /manager
         args:
-        - --leader-elect
+        - "--leader-elect"
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
-        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
+        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true},ClusterTopology=${CLUSTER_TOPOLOGY:=true}"
+        - "--namespace=$(WATCH_NAMESPACE)"
```
anmazzotti, line 29: These should use curly brackets `{`, I'm not sure it will count as an env variable for clusterctl otherwise.

apoorvajagtap: I tried the following pattern to maintain consistency earlier:
```
- "--namespace=${WATCH_NAMESPACE:=}"
- "--concurrency=${CONCURRENCY_NUMBER:=10}"
```
expecting an empty string to be picked up if `WATCH_NAMESPACE` is not exported. But, this has been leading to e2e test failures with `WATCH_NAMESPACE` unset (works well for concurrency though).
So, I opted for the native k8s way for env, which seems to work.

bootstrap/config/manager/manager.yaml
```diff
@@ -21,11 +21,13 @@ spec:
...
         - /manager
         args:
-        - --leader-elect
+        - "--leader-elect"
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
-        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
+        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true},ClusterTopology=${CLUSTER_TOPOLOGY:=true}"
+        - "--namespace=${WATCH_NAMESPACE:=''}"
```
anmazzotti, line 29: I think `""` is the empty string, not sure it will work with single quotes.

bootstrap/config/manager/manager.yaml
```diff
@@ -21,11 +21,16 @@ spec:
...
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
-        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
+        - "--feature-gates=MachinePool=${EXP_MACHINE_POOL:=true},ClusterTopology=${CLUSTER_TOPOLOGY:=true}"
+        - "--namespace=$(WATCH_NAMESPACE)"
+        - "--concurrency=${CONCURRENCY_NUMBER:=10}"
+        env:
+          - name: WATCH_NAMESPACE
+            value: ${WATCH_NAMESPACE:=""}
```
furkatgofurov7, line 33: Where are these feature flags defined? I don't seem to find them anywhere..

furkatgofurov7: correct, but I don't like the fact that it is just defined as an arg and not defined/used anywhere, like in e2e/conf.yaml we have bunch of these set like https://github.com/rancher/cluster-api-provider-rke2/blob/5ae52a57956daf32b39d9f949b7650273e341a9b/test/e2e/config/e2e_conf.yaml#L116 for example

alexander-demicev (approved):

bootstrap/config/manager/manager.yaml
anmazzotti: That I added because I was struggling with logs iirc. 
I don't think we need to test argument mapping, I'd trust clusterctl to work fine and they are already implicitly tested with the defaults imho

furkatgofurov7 (approved):

PR #697, opened 2025-07-11
Implement control plane load balancer exclusion

### LINKED ISSUE #696: need to have RKE2 provider set exclude-from-external-load-balancers on CP nodes that are being deleted
## Problem statement

When using RKE2 with MetalLB, with MetalLB speakers running on control-plane nodes, we would need MetalLB to gracefully stop advertising virtual IPs; or else we may enter a state where metallb still advertises a virtual IP but has lost the ability to reliably reach kubernetes API and cannot know that it should relinquish the VIP.

## Non solution

Metallb 0.14.9 had an evolution which would have let us obtain exactly this behavior (https://github.com/metallb/metallb/pull/2470), but this evolution was unfortunately removed in 0.15.0 (https://github.com/metallb/metallb/pull/2715).

## Proposed solution

**Credits to @zioc for the base idea, to @anmazzotti for suggesting the opt-in aspect**

The proposed solution is based on the fact that [metallb honors](https://github.com/metallb/metallb/pull/2073) the [exclude-from-external-load-balancers well-known label](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-exclude-from-external-load-balancers).

The behavior would be the following: cabprke2 would have an optional behavior (based on some `rke2.controlplane.cluster.x-k8s.io/load-balancer-exclusion` new annotation that would be introduced for this purpose) that would consist in labeling a Node with the [exclude-from-external-load-balancers well-known label](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-exclude-from-external-load-balancers) on Machine Node drain, more precisely in a pre-drain CAPI Machine hook (doing this labeling later, e.g. in pre-terminate would probably be too late, possibly not giving metallb sufficient time to react before etcd is killed).  Once the node would be labelled, metallb would gracefully stop advertising the virtual IP from the speaker running on that node.

## Background Context

The idea described here started from a discussion with @anmazzotti and @zioc around metallb and etcd membership removal.

* thanks to https://github.com/rancher/cluster-api-provider-rke2/pull/684, there will now be a clean way to have cabprke2 remove etcd membership 
* Sylva, which can't yet use this unreleased evolution is using a custom build of cabprke2 which skips the etcd membership removal code -- we of course don't want to do this in the long tun
* until today we had in mind that since metallb 0.14.9 we could consider that metallb behavior was fine during node drain/deletion (but as said above since the revert  in 0.15.x, it is **not**)
* while testing this evolution in Sylva, we noticed problems of metallb VIP reachability, which led to reconsider the part that metallb plays during CP node rolling updates

/cc @hardys

anmazzotti:
This PR introduces a `RKE2ControlPlane` annotation: `rke2.controlplane.cluster.x-k8s.io/load-balancer-exclusion`.

This annotation will apply the `pre-drain.delete.hook.machine.cluster.x-k8s.io/rke2-lb-exclusion` annotation on all control plane Machines.
The purpose of the hook is to apply the `node.kubernetes.io/exclude-from-external-load-balancers: "true"` label on the underlying Node, giving the external load balancer occasion to stop advertising it.

The `RKE2ControlPlane` annotation can be toggled, to enable or disable this feature accordingly.

Fixes #696

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


zioc:
I've just raised an MR in sylva to test how it works: https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/4874

(The last tests of [release 0.17.1](https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/4763) and [build of PR 684](https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/4766) were showing multiple failures in upgrades, we'll see if that PR helps)

alexander-demicev:
@anmazzotti I added do not merge label, feel free to remove it once we can merge the PR

controlplane/internal/controllers/lifecycle_hook.go
```diff
@@ -0,0 +1,378 @@
...
+	if deletingMachine == nil {
+		log.V(5).Info("Waiting on other machines to be deleted.")
+
+		return ctrl.Result{RequeueAfter: deleteRequeueAfter}, nil
+	}
+
+	log = log.WithValues("Machine", klog.KObj(deletingMachine))
+	ctx = ctrl.LoggerInto(ctx, log)
+
+	if controlPlane.Machines.Len() < 2 {
```
zioc, line 128: Could it even happen? Draining and deleting the last CP node will inevitably lead to cluster failure...

This extra check won't hurt in any case.

anmazzotti: We do allow users to break their clusters, yes. :D
It will be fixed: https://github.com/rancher/cluster-api-provider-rke2/issues/701

The extra check is needed now and I think it will be good to stay just in case.

PR #698, opened 2025-07-11
Add ConfigMap support as FileSource

### LINKED ISSUE #565: Add ConfigMap support as FileSource for write_files configuration
**Describe the solution you'd like**:

Add support for ConfigMaps as a FileSource in write_files configuration for cloud-init. Similar to the existing SecretFileSource functionality, we need the ability to reference ConfigMap data when generating files during initialization. This will allow getting file content from either Secrets or ConfigMaps depending on the use case.

Currently file content can only be sourced from Secrets, which is limiting for non-sensitive configuration data
ConfigMaps are the standard Kubernetes way to handle non-sensitive configuration
This will provide more flexibility in managing configuration data
Users can better separate sensitive data (in Secrets) from regular configuration (in ConfigMaps)

The implementation should maintain backwards compatibility with existing SecretFileSource
Only one source type (Secret or ConfigMap) should be allowed at a time for a given file
This enhancement follows the standard Kubernetes pattern of using ConfigMaps for configuration data
Documentation will need to be updated to reflect the new ConfigMap support

valaparthvi:
kind/feature
1. Add ConfigMap support as a FileSource
2. `Secret` and `ConfigMap` both will be marked as optional, since either `ConfigMap` or `Secret` must be provided when fetching contentFrom file. A new `FileSourceRef` structure is added, replacing `SecretFileSource`.
3. Add kubebuilder validation to ensure either one is provided if fetching contentFrom FileSource, and only one is provided. ~~Add unit tests for the same.~~
4. Add e2e test for this.
5. Add documentation

Fixes #565 

**Checklist**:

- [x] includes documentation
- [x] adds unit tests


bootstrap/api/v1alpha1/rke2config_types.go
```diff
@@ -328,6 +333,18 @@ type SecretFileSource struct {
 	Key string `json:"key"`
 }
 
+// ConfigMapFileSource adapts a ConfigMap into a FileSource.
```
alexander-demicev, line 336: We don't add new features to previous API versions, this can be removed from v1alpha1.

bootstrap/api/v1beta1/rke2config_types.go
```diff
@@ -338,6 +343,18 @@ type SecretFileSource struct {
 	Key string `json:"key"`
 }
 
+// ConfigMapFileSource adapts a ConfigMap into a FileSource.
+//
+// The contents of the target ConfigMap's Data field will be presented
+// as files using the keys in the Data field as the file names.
+type ConfigMapFileSource struct {
```
alexander-demicev, line 350: ConfigMapFileSource and SecretFileSource are identical, maybe we can have single `FileSourceRef` structure for both?

bootstrap/api/v1beta1/rke2config_webhook.go
```diff
@@ -158,10 +158,34 @@ func (s *RKE2ConfigSpec) validate(pathPrefix *field.Path) field.ErrorList {
 
 	allErrs = append(allErrs, s.validateIgnition(pathPrefix)...)
 	allErrs = append(allErrs, s.validateRegistries(pathPrefix)...)
+	allErrs = append(allErrs, s.validateFileSource(pathPrefix)...)
 
 	return allErrs
 }
 
+// validateFileSource ensures that either ConfigMap or Secret FileSource is provided; it raises error if both or none are provided
+func (s *RKE2ConfigSpec) validateFileSource(pathPrefix *field.Path) field.ErrorList {
```
alexander-demicev, line 167: out of the scope for this PR but we might want use CEL instead of admission webhooks

bootstrap/internal/controllers/rke2config_controller.go
```diff
@@ -548,28 +548,50 @@ func (r *RKE2ConfigReconciler) generateFileListIncludingRegistries(
 
 	for _, file := range scope.Config.Spec.Files {
 		if file.ContentFrom != nil {
-			scope.Logger.V(5).Info("File content is coming from a Secret, getting the content...")
+			if file.ContentFrom.Secret.DeepCopy() != nil {
```
alexander-demicev, line 551: what is the reason to use deepcopy here?

alexander-demicev (changes requested):
thanks, overall looks great, I have a couple of small comments

bootstrap/api/v1beta1/rke2config_types.go
valaparthvi: That would differ from the other bootstrap providers that have `SecretFileSource`, if you think it will be okay, I can add a `FileSourceRef` with something like:
```
type FileSourceRef struct {
   	// Name of the source in the RKE2BootstrapConfig's namespace to use.
	Name string `json:"name"`

	// Key is the key in the source's data map for this value.
	Key string `json:"key"`

	// Type is the type of the source, this value can either be secret or configMap.
        Type string `json:"type"` 
}
```

`Type` would help determine which resource to fetch the data from.

bootstrap/internal/controllers/rke2config_controller.go
valaparthvi: `file.ContentFrom.Secret` is not a pointer, and I was not sure which version of `SecretFileSource` to use to check if it is empty or not, `DeepCopy()` helped simplify the check. I did think of using `file.ContentFrom.Secret.Name != ""` or something similar, but then decided to go ahead with `DeepCopy()` check.

bootstrap/api/v1beta1/rke2config_types.go
alexander-demicev: I meant something like this:
```golang
type FileSource struct {
	// SecretFileSource represents a secret that should populate this file.
	//+optional
	Secret FileSourceRef `json:"secret"`

	// ConfigMapFileSource represents a ConfigMap that should populate this file.
	//+optional
	ConfigMap FileSourceRef `json:"configMap"`
}

type FileSourceRef struct {
	// Name of the secret in the RKE2BootstrapConfig's namespace to use.
	Name string `json:"name"`

	// Key is the key in the secret's data map for this value.
	Key string `json:"key"`
}
```
wdyt?

valaparthvi: I modified this. I am now seeing the following error since these are undefined functions, but I am not a 100% sure if I should define them manually or add some conversion logic to `RKE2Config`'s `ConvertTo` and `ConvertFrom` methods in the `bootstrap/api/v1alpha1/conversion.go` file and then run `make generate` which might resolve these errors.

```
# github.com/rancher/cluster-api-provider-rke2/bootstrap/api/v1alpha1 [github.com/rancher/cluster-api-provider-rke2/bootstrap/api/v1alpha1.test]
./zz_generated.conversion.go:76:10: undefined: Convert_v1beta1_FileSource_To_v1alpha1_FileSource
./zz_generated.conversion.go:305:13: undefined: Convert_v1beta1_FileSource_To_v1alpha1_FileSource
./zz_generated.conversion.go:320:12: undefined: Convert_v1alpha1_SecretFileSource_To_v1beta1_FileSourceRef
./zz_generated.conversion.go:332:12: undefined: Convert_v1beta1_FileSourceRef_To_v1alpha1_SecretFileSource
```

Should I define these functions manually?

alexander-demicev (approved):
Thanks, all looks good

alexander-demicev (approved):
thanks for working on this

PR #794, opened 2026-01-13
:sparkles: Bump to CAPI v1.11.5 and adopt `v1beta2` contract

### LINKED ISSUE #732: Bump CAPI to v1.11.0
Migration guide: https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.10-to-v1.11

Also, we need to take care of adding `kind: Metadata` to all metadata manifests in the repo, xref: https://github.com/kubernetes-sigs/cluster-api/pull/12242.

salasberryfin:
kind/feature

This bumps the provider to CAPI v1.11.5 and adopts the new `v1beta2` CAPI contract.

As control plane/bootstrap providers are highly coupled to core CAPI objects, there's no middle ground or multi-step adoption of CAPI v1.11 like other infrastructure providers are doing. This requires us to create a fully new CAPRKE2 API version `v1beta2` in which we can integrate all of the breaking changes that come with the migration.

The main changes that are addressed by this PR (please, refer to the [migration guide](https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.10-to-v1.11) for a full list):
- The transition to the new K8s aligned conditions using `metav1.Conditions` types and the new condition semantic has been completed. This requires a complete rebuild of the existing `Status` definitions. For the time being, and inline with the upstream recommendation, the CAPI-specific conditions are kept until `v1beta1` is completely dropped, but are moved to `status.deprecated.v1beta1`.
- The `spec.controlPlaneRef` and `spec.infrastructureRef` fields are now using `ContractVersionedObjectReference` type instead of `corev1.ObjectReference`. This means that fields like `namespace`, `resourceVersion`, `apiVersion` are no longer available and `apiGroup` must be used instead. With this change, the object version is not explicitly specified in the spec and will be read from the corresponding CRD

There are many other changes but it is impossible to list them all in a PR description. These two are probably the ones that require the most changes, including CRD, controller logic and API conversion.

The goal of this first iteration on CAPI v1.11 support is to have a fully functional provider based on the following E2E suites:
- [x] Workload cluster creation.
- [x] Cluster Class provisioning.
- [x] External Datastore.
- [x] In-place propagation.
- [x] RCP remediation
- [x] Provider upgrade: it now requires dealing with API version changes and this creates an extra layer of complexity for validating resource health.

Tests that are skipped:
- Bootstrap & Pivot.

**Which issue(s) this PR fixes**:
Fixes #732 

Built on top of #793 and #798.

After this PR is merged, there will be follow-ups to improve and streamline the implementation, including better testing, condition management and specific API definitions.

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,3 +31,29 @@ jobs:
           name: e2e-artifacts
           path: _artifacts
           if-no-files-found: ignore
+  e2e-tests-upgrade:
+    runs-on: runs-on,runner=4cpu-linux-x64,run-id=${{ github.run_id }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - name: Install Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+      - name: Setup kind
+        uses: helm/kind-action@v1.13.0
+        with:
+          install_only: true
+      - name: Run E2E tests
+        env:
+          GINKGO_LABEL_FILTER: upgrade
+        run: make test-e2e
+      - name: Archive artifacts
+        if: always()
+        uses: actions/upload-artifact@v6.0.0
+        with:
+          name: e2e-upgrade-artifacts
+          path: _artifacts
+          if-no-files-found: ignore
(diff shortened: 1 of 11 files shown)


alexander-demicev (approved):
This is a huge amount of work, thanks a lot for taking care of the update!

alexander-demicev (approved):

PR #866, opened 2026-03-25
Add support for in-place update contracts

alexander-demicev:

This PR add support for in-place update contracts. It requires some internal CAPI functions to be made public but that's likely not going to happen anytime soon https://github.com/kubernetes-sigs/cluster-api/pull/13507. I added a CAPI package that includes all required code and keeps it separated. We can later drop the package as we make functions public.

The carried code from CAPI is added with a separate commit to make reviews easier.

Fixes https://github.com/rancher/cluster-api-provider-rke2/issues/865

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


furkatgofurov7:
@alexander-demicev looks like `capi-controller-manager` is not coming up in e2e tests, maybe feature gate introduction broke it?

controlplane/config/manager/manager.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         - "--diagnostics-address=${CAPRKE2_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPRKE2_INSECURE_DIAGNOSTICS:=false}"
         - "--v=${CAPRKE2_DEBUG_LEVEL:=0}"
-        - "--concurrency=${CONCURRENCY_NUMBER:=10}"
```
anmazzotti, line 28: I miss why we lost the configurable concurrency variable here.

controlplane/main.go
```diff
@@ -244,11 +271,49 @@ func setupReconcilers(ctx context.Context, mgr ctrl.Manager) {
 		os.Exit(1)
 	}
 
+	var runtimeClient runtimeclient.Client
+
+	if feature.Gates.Enabled(feature.InPlaceUpdates) {
```
anmazzotti, line 276: Would be sweet to have an info log here saying InPlaceUpdate is enabled.

controlplane/internal/controllers/inplace_canupdatemachine.go
```diff
@@ -0,0 +1,362 @@
...
+	extensionHandlers, err := r.RuntimeClient.GetAllExtensions(ctx, runtimehooksv1.CanUpdateMachine, machine)
+	if err != nil {
+		return false, err
+	}
+	// Machine cannot be updated in-place if no CanUpdateMachine extensions are registered.
+	if len(extensionHandlers) == 0 {
+		return false, nil
+	}
+
+	if len(extensionHandlers) > 1 {
```
anmazzotti, line 73: I don't understand when/how this can happen, also considering `canExtensionsUpdateMachine` is already set to loop through multiple handlers.

anmazzotti (dismissed):
I left some nitpicks and questions, but this looks good to me.
Also thank you a lot for splitting the library commit.

furkatgofurov7 (dismissed):

controlplane/config/manager/manager.yaml
alexander-demicev: thanks for catching, it was removed by mistake

controlplane/internal/controllers/inplace_canupdatemachine.go
alexander-demicev: I think the reason is to liimit the amount of supported hooks to 1 for now but with option to support more if needed. This code follows kubeadm provider pattern.

anmazzotti (approved):

furkatgofurov7 (approved):

PR #880, opened 2026-04-14
fix: add checksum validation for curl/wget downloads

mbologna:
kind/security

- download-rke2-artifacts.sh: download sha256sum file first and verify each artifact against it; also harden the script with strict shell options and proper quoting
- hack/ensure-kubectl.sh: pin kubectl v1.27.0 SHA-256 checksums for linux/darwin × amd64/arm64; verify binary after download. Pin krew to v0.5.0 (was 'latest') with per-arch checksums and verify archive before extracting. Add Renovate annotations so both version and digests are kept up-to-date automatically.
- image-builder/scripts/opensuse/bootstrap.sh: pin AWS CLI to v2.34.30 with a versioned URL and SHA-256 verification; download the RKE2 sha256sum file first and verify each artifact (rke2-images, rke2 tarball) immediately after download.
- scripts/install-mdbook.sh: download to a temp file, verify the SHA-256 checksum, then extract – eliminating the unsafe curl | tar pipe pattern. Add Renovate annotations.

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/examples/templates/docker/air-gapped/image-building/download-rke2-artifacts.sh
+++ b/examples/templates/docker/air-gapped/image-building/download-rke2-artifacts.sh
@@ -3,24 +3,43 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-: "${RKE2_VERSION:=v1.26.0+rke2r1}"
+# renovate: datasource=github-release-attachments depName=rancher/rke2
+RKE2_VERSION="v1.26.0+rke2r1"
 
-ARCH="${ARCH:-amd64}"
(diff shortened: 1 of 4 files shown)


salasberryfin (approved):
Thanks @mbologna.

I think these changes are good from a security perspective.

We probably need to do some housekeeping on the versions used in these scripts (i.e. `kubectl`, `rke`, etc.) but this is for a follow-up.

anmazzotti (approved):

PR #906, opened 2026-05-19
fix: properly format IPv6 addresses in server URLs

Kristian-ZH:
When provisioning nodes with IPv6 control plane endpoints, the bootstrap provider was generating invalid server URLs in the RKE2 configuration.

Invalid format:  https://fd12:3456:890b::250:6443
Correct format:  https://[fd12:3456:890b::250]:6443

This fix adds a formatServerURL() helper function that:
- Detects IPv6 addresses using net.ParseIP()
- Wraps IPv6 addresses in brackets per RFC 3986
- Leaves IPv4 addresses and hostnames unchanged

The fix is applied to all three server URL generation points:
- Init control plane config
- Join control plane config
- Worker node config

### DIFF
--- a/bootstrap/internal/controllers/rke2config_controller.go
+++ b/bootstrap/internal/controllers/rke2config_controller.go
@@ -22,6 +22,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"strconv"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -62,7 +64,6 @@ import (
 const (
 	filePermissions  string = "0640"
 	registrationPort int    = 9345
-	serverURLFormat  string = "https://%v:%v"
 	tokenPrefix      string = "-token"
 )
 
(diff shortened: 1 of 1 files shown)


bootstrap/internal/controllers/rke2config_controller.go
```diff
@@ -66,6 +67,23 @@ const (
...
+func formatServerURL(host string, port int) string {
+	// Check if the host is an IPv6 address by attempting to parse it
+	ip := net.ParseIP(host)
+	if ip != nil && ip.To4() == nil {
+		// IPv6 address - wrap in brackets
+		return fmt.Sprintf("https://[%s]:%d", host, port)
+	}
+	// IPv4 address or hostname - no brackets needed
+	return fmt.Sprintf(serverURLFormat, host, port)
+}
```
hardys: Instead of using a new function you could just use https://pkg.go.dev/net#JoinHostPort which does the same thing?

hardys (changes requested):

bootstrap/internal/controllers/rke2config_controller.go
Kristian-ZH: thanks, addressed!

anmazzotti (approved):

hardys (approved):
lgtm, thanks!

furkatgofurov7 (approved):

furkatgofurov7:
@Kristian-ZH hi, thanks for the change, however, the repository has required settings so that all commits must have verified signatures, otherwise we can't merge it. Please sign your commit, the change LGTM!

salasberryfin (approved):

Issue #908, opened 2026-05-20
Add E2E coverage


salasberryfin:
# Description

Leverage the update extension to write e2e tests.

furkatgofurov7:
#910

PR #910, opened 2026-05-21
Add in-place updates e2e suite

### LINKED ISSUE #907: Write a mock update extension for CAPRKE2
# Description

Use recent changes in Kubeadm as reference for implementing an update extension for CAPRKE2

alexander-demicev:

This PR adds e2e suite and test extension for in-place updates, I also uncovered some issues with the implementation. All code mostly follows kubeadm patterns. 

Fixes #907, #908 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


pkg/rke2/desiredstate/desired_state.go
```diff
@@ -0,0 +1,262 @@
...
+		// Creating a new machine
+		machineName = names.SimpleNameGenerator.GenerateName(rcp.Name + "-")
+		version = rcp.GetDesiredVersion()
+
+		serverConfig, err := json.Marshal(rcp.Spec.ServerConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal server configuration: %w", err)
+		}
+
+		annotations[controlplanev1.RKE2ServerConfigurationAnnotation] = string(serverConfig)
```
anmazzotti, line 102: I miss why we propagate the serverConfig to the machine as annotation.
Is this used anywhere else? This could be quite long since it has many nested fields.

test/extension/handlers/inplaceupdate/handlers.go
anmazzotti, line 1: Asking out of ignorance. Is this the in-place update webhook implementation? Will this be running within the same caprke2 controller or require its own deployment?
Should it already live in some `exp`(erimental) subfolder and be deployed depending on a feature gate?

alexander-demicev: It will be a separate deployment, same as any other CAPI extension. This extension is for testing contracts only, the real one can live anywhere.

pkg/rke2/desiredstate/desired_state.go
alexander-demicev: It used here https://github.com/rancher/cluster-api-provider-rke2/blob/main/pkg/rke2/machine_filters.go#L104 to compare with `ServerConfig` for deciding if machine needs a rollout, your concern is valid but the pattern is already being used.

anmazzotti: Thank you, I missed that.
Ok it makes sense. 

I was wondering if we can just compare (and propagate) the hash of the (json serialized) object, but that also has its issues as we can't make distinctions between supported fields or not.

anmazzotti (approved):
Nice work, thank you!

furkatgofurov7 (approved):
Thanks a lot, nice work!

alexander-demicev:
@furkatgofurov7 @anmazzotti can someone restart the lint job?

furkatgofurov7:
cc @salasberryfin

controlplane/internal/controllers/rke2controlplane_controller.go
```diff
@@ -119,6 +124,7 @@ type RKE2ControlPlaneReconciler struct {
 // +kubebuilder:rbac:groups="infrastructure.cluster.x-k8s.io",resources=*,verbs=get;list;watch;create;patch;delete
 // +kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;list;watch
 // +kubebuilder:rbac:groups=runtime.cluster.x-k8s.io,resources=extensionconfigs,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
```
salasberryfin, line 127: I'm curious to know if this is something we need specifically for in-place updates or if this was missing from the provider.

yiannistri (approved):
Thank you

controlplane/internal/controllers/rke2controlplane_controller.go
furkatgofurov7: This does not seem to be a blocker, but rather a question AFAIU and even it is we can follow-up, so I am going ahead with adding this PR to the merge queue before doing a new minor release

alexander-demicev: `GetAllExtensions()` does a get on namespaces internally https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/runtime/client/client.go#L766

alexander-demicev: Yes, it's in-place updates specific

PR #969, opened 2026-07-30
Add ingress controller API field

alexander-demicev:

Add ingress controller API field, implementation follows RKE2 docs https://docs.rke2.io/reference/server_config#networking

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controlplane/api/v1beta1/rke2controlplane_types.go
+++ b/controlplane/api/v1beta1/rke2controlplane_types.go
@@ -202,6 +202,13 @@ type RKE2ServerConfig struct {
 	//+optional
 	CNIMultusEnable bool `json:"cniMultusEnable,omitempty"`
 
+	// IngressController describes the ingress controllers to deploy, one of none, traefik, ingress-nginx.
+	// All selected ingress controllers are deployed, and the first value is set as the default ingress
+	// class. A single value of none disables all packaged ingress controllers.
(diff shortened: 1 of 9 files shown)


anmazzotti (approved):
I miss where `All selected ingress controllers are deployed`. I find this part of the comment misleading.

In the RKE2 [docs](https://docs.rke2.io/reference/server_config#networking) it says instead:
```
Ingress Controllers to deploy, one of none, traefik, ingress-nginx; the first value will be set as the default ingress class
```
I also find this a bit confusing, because it says **one** value, but then mentions the **first**? 

However looking at the KDM this seems defined as an array, so all good I guess.
https://github.com/rancher/kontainer-driver-metadata/blob/release-v2.15/data/data.json#L54180C1-L54187C8

Not sure if the comment can be adjusted, I don't know if I should expect **all** controllers to be deployed or only the **first** one.

Maybe I'm overlooking this and actually they are all deployed and only the first one is set to default, which is exactly what the description says. I can't read, all good. Thank you for the PR!

alexander-demicev:
@anmazzotti I also got a bit confused, I think usually you'd choose one of three options but since it's an array I used slice for parity

furkatgofurov7 (approved):

PR #989, opened 2026-08-07
chore: bump CAPI to 1.14.0

### LINKED ISSUE #957: Bump to CAPI v1.14
# Description

CAPI v1.14.0 will be released on August 11th.

Update the provider and create a new minor release.

yiannistri:

Fixes #957 

Bumping to cluster-api `v1.14.0` needed:
- import new `cluster-api/api` module
- fix existing imports `/exp/runtime/catalog` -> `/api/runtime/catalog`
- fix `cache.New` calls to pass a context

Notable change: in 1.13, the in-place e2e test was setting the Kubernetes version and asserted that no rollout happened. This is no longer possible in 1.14, because a [newly added check](https://github.com/kubernetes-sigs/cluster-api/pull/13871) does not set the machine as UpToDate, if `spec.version` is different from the `status.nodeInfo.kubeletVersion`, something that a fake in-place extension cannot perform. I would expect an e2e test to cover this scenario for the real in-place extension.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


test/e2e/e2e_in_place_update_test.go
```diff
@@ -249,30 +249,8 @@ var _ = Describe("In-place update via Runtime Extension", Label(DefaultTestsLabe
 			}, e2eConfig.GetIntervals(specName, "wait-control-plane")...).Should(Succeed())
 		}
 
-		By("Triggering an in-place Kubernetes version upgrade")
```
furkatgofurov7, line 252: Is there a reason why we have removed this code?

test/e2e/config/e2e_conf.yaml
```diff
@@ -40,6 +40,17 @@ providers:
             new: "imagePullPolicy: IfNotPresent"
           - old: "--leader-elect"
             new: "--leader-elect=false"
+      - name: "v1.14.0"
```
furkatgofurov7, line 32: We can remove the v1.12.8 entry in 2 places

furkatgofurov7 (commented):
Thanks for the PR @yiannistri 

I have few suggestions:
- starting from CAPI v1.15 Docker* API resources will be [deprecated](https://github.com/kubernetes-sigs/cluster-api/releases/#release-v1.14.0) (please see **Deprecation Warning** section) and it would be good to move to Dev* API resources in this release bump. There is a (partial) prior art in https://github.com/rancher/turtles/pull/2690/changes
- we need to check if we are still consuming 1beta1 API version of core Cluster API, CABPK and KCP and move to v1beta2 if possible (can also be tracked/worked outside of this change)

test/e2e/e2e_in_place_update_test.go
yiannistri: In 1.14, a check was [added](https://github.com/sbueringer/cluster-api/blob/560d4acf507bc7cac34b2da449fa5cd53eaeb149/core/reconcilers/machine/machine_controller_status.go#L726-L735) that prevents setting the machine as UpToDate, when the machine's spec version does not match the machine's kubelet version in status. Since the fake extension does not _really_ update the nodes' version, the cluster never becomes ready and the test times out. I would expect this exact scenario to be tested with a real in-place upgrades extension.

furkatgofurov7 (dismissed):
Great work @yiannistri

yiannistri:
Added https://github.com/rancher/cluster-api-provider-rke2/issues/1002 as a follow up based on https://github.com/rancher/cluster-api-provider-rke2/pull/989#pullrequestreview-4958718433

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/rancher/cluster-api-provider-rke2
 
-go 1.25.10
+go 1.26.0
```
salasberryfin: Can we bump to 1.26.6 for the latest security patches?

salasberryfin (approved):

furkatgofurov7 (approved):

